Issue #5, opened 2022-07-12
GRPC packet ingestion


macpie:
Helium gateways speak gRPC to Helium LoRaWAN Network Servers (LNS)

- [x] #10
  + GitHub-CI is successful with grpc builds
- [x] Define new protobuf:  (Thanks, Pierre!)  See [packet_router.proto](https://github.com/helium/proto/blob/macpie/packet_router/src/service/packet_router.proto) in https://github.com/helium/proto/pull/157
- [x] Listen for inbound gRPC connections (https://github.com/helium/helium-packet-router/pull/25)

EDIT: unnecessary former tasks or accommodated through other means:
- Tee raw gRPC traffic to local file:
  + Required as an input to our Packet Verify service (lives outside of HPR or LNS)
  + Accommodates replaying traffic for debugging
  + Accommodates live traffic going to next Release Candidate but with all safeguards of Production environment
- Forward to `router` as gRPC (https://github.com/helium/helium-packet-router/pull/30)
- Accommodate load-balancer sessions with id/token in gRPC values or headers

Notes:

- Inbound traffic from gateway-rs will be gRPC flowing through a load-balancer
   + AWS app layer ELB is gRPC aware
- Handling of large scale traffic volumes will be deferred to load-balancer
   + e.g., Let LB maintain large socket listen queue because setup/shutdown of sockets from LB to packet-router across AWS backbone should be relatively cheap
   + Therefore, nothing special to be done for tuning Linux kernel or Erlang/BEAM
- This ticket is just for the Listening portion of gRPC
   + The companion use of `http2_client` for accommodating sending to Class C devices (downlink without an uplink initiating the request) will be implemented as a separate effort

macpie:
This should probably be split up

dpezely:
> This should probably be split up

Updated description with task list / TODO items.  We can further break out items as needed later.

dpezely:
Where `router` used its own Erlang behavior, `helium_router_bhvr`,
packet-router functions more like yet another intermediate gateway.
Therefore, its behavior will be `gateway_bhvr` in terms of gRPC's autogen components.

michaeldjeffrey:
This branch https://github.com/helium/helium-packet-router/pull/25 and this branch https://github.com/helium/helium-packet-router/pull/27 both accept packets over grpc when running a local gateway-rs on this branch https://github.com/helium/gateway-rs/pull/289.

[...]

PR #9, opened 2022-07-14
Add boilerplate config and code for gRPC and protobufs

dpezely:
This PR addresses first task of [issue #5](https://github.com/helium/helium-packet-router/issues/5) for Helium `packet-router` which functions as an intermediary gateway for LoRaWAN.

Notes: 

- This is **not a full router / LNS**
- For implementing the behavior, `helium_gateway_bhvr`, code was replicated from [sibyl](https://github.com/helium/sibyl) strictly for purposes of exercising the config and dependencies
   + Code was copied for merely attaining a clean compile via `make compile` (or simply, `make`)
   + Specifically, `helium-packet-router/src/grpc/helium_packet_router_service.erl` was replicated from `router`
   + This code **has NOT been executed**
- This is also to establish a baseline from which other branches can easily extend
   + Let's ensure that `rebar.config`, `sys.config`, etc. within this PR are minimal and clean
   + Let's keep rebar.config mostly alphabetical for its sequence of `plugins`, `prehooks`, etc.
- This is **without tests** of any kind but will be added in a future PR
   + because this PR is mostly about rebar.config, etc.
- Modules called from `src/grpc/helium_packet_router_service.erl` are omitted from this PR
  + Those will be in the next PR

**Updates:**

- The version of `.github/workflows/tests.yml` within `main` branch omits `make grpc` before `./rebar3 compile`, so this PR's build fails in CI
  + The fix is included here with commit 4ff36a4
  + But it apparently only gets applied when committed to `main` branch-- not the PR's own branch
  + *sigh*
- An early commit used same Erlang behavior from `sibyl` but changed to that from `router` for simplicity

@andymck I followed your trail of breadcrumbs through various repos such as `gateway-rs`, `blockchain-core`, `miner` and of course `sibyl`.  Thank you for all your earlier work on each of those.  If there's any legacy config or code that doesn't need to be ported here for `packet-router`, please let us know!


rebar.config
```diff
@@ -1,17 +1,38 @@
...
+
 {plugins, [
+    {grpcbox_plugin,
+        {git, "https://github.com/novalabsxyz/grpcbox_plugin.git", {branch, "master"}}},
     {rebar3_format, "1.2.1"},
     {erlfmt, "1.0.0"}
 ]}.
 
+{pre_hooks, [
+    {compile, "make grpc"},
```
michaeldjeffrey, line 30: Is there a way we can not have this pre_hook?
It recompiles grpc when you do anything with rebar3.

rebar.config
```diff
@@ -1,17 +1,38 @@
-{deps, []}.
+{deps, [
+    {erlang_lorawan, ".*",
+        {git, "https://github.com/helium/erlang-lorawan.git", {branch, "master"}}},
+    {grpc_client,
+        {git, "https://github.com/Bluehouse-Technology/grpc_client.git", {branch, "master"}}},
+    {grpcbox, {git, "https://github.com/novalabsxyz/grpcbox.git", {branch, "master"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
```
ruttenberg, line 7: I don't think it is a good idea to specify a dependency via branch name. This practice leaves us open to accidentally absorbing library changes that weren't intended for release.
Yes, I know that the change won't be absorbed unless _rebar.lock_ is updated, but that has two weaknesses:

1.  The risk of accidentally committing the change is increased when other dependency changes are in _rebar.lock_.
2. Nothing is communicated about the  _intent_ of the change. When a tag (or at minimum, a commit hash) is used, it is clear that the change was consciously made.

We don't live in a world where mistakes are never committed to significant branches, nor where our testing schemes are likely to save us from not thinking about these changes.

rebar.config
dpezely: Yes, there is a way.  It was already accounted for via `Makefile` such that this `pre_hook` was redundant.

e.g., within Makefile: `compile: | $(grpc_services_directory)`

Fixed in commit 5531bd9.

Makefile
```diff
@@ -1,15 +1,17 @@
-.PHONY: compile clean test rel run docker-build docker-test docker-run
+.PHONY: compile clean test rel run grpc docker-build docker-test docker-run
+
+grpc_services_directory=src/grpc/autogen
 
 REBAR=./rebar3
 
-compile: |
-	$(REBAR) compile 
+compile: | $(grpc_services_directory)
```
andymck, line 7: on miner we removed this pre requisite approach of handling grpc and instead used rebar prehooks.  Be worth checking that setup out.  Keeps things cleaner and more maintainable

src/grpc/helium_packet_router_service.erl
```diff
@@ -0,0 +1,45 @@
+-module(helium_packet_router_service).
```
andymck, line 1: noting it here for reference, this handler was one of the first grpc handlers added and its approach  with the 'receive' loop is maybe a little raw.

The newer handlers implement handle_info msgs and as such are a somewhat more of an otp style approach.  An example of those newer handlers can be seen here:

https://github.com/helium/blockchain-core/blob/master/src/grpc/blockchain_grpc_sc_server_handler.erl

This router handler is obviously a lot simpler and like i say nothing wrong with it on its own but maybe be better to follow the newer pattens of the later handlers especially if this is going to be the basis of other handlers

config/grpc_client_gen.config
```diff
@@ -0,0 +1,34 @@
+{plugins, [
```
andymck, line 1: note: client config only required if you plan on adding grpc tests and need to execute client behaviour

rebar.config
```diff
@@ -1,17 +1,36 @@
...
+    {lager, "3.9.2"}
+]}.
 
 {erl_opts, [
     debug_info,
+    {parse_transform, lager_transform},
     warnings_as_errors
 ]}.
 
+{grpc, [
```
andymck, line 19: i dont think you need this section

rebar.config
```diff
@@ -1,17 +1,36 @@
...
+    {gpb_opts, [{module_name_suffix, "_pb"}]}
+]}.
+
 {plugins, [
+    {grpcbox_plugin, {git, "https://github.com/novalabsxyz/grpcbox_plugin.git"}},
     {rebar3_format, "1.2.1"},
     {erlfmt, "1.0.0"}
 ]}.
 
+{pre_hooks, []}.
```
andymck, line 30: as per previous comment, better to call the grpc makefile target here from a compile pre hook.  likewise for a clean prehook.  see https://github.com/helium/miner/blob/master/rebar.config#L61

Makefile
dpezely: Apparently, one side-effect of using rebar3's `pre_hooks` is that grpc modules get regenerated too frequently.

We originally had `pre_hooks` consistent with miner in the first commit of this branch but requested to be removed from `rebar.conf`, so then `make compile` for full builds and `rebar compile` for rapid iterations.

Is there a flaw in that reasoning?

config/grpc_client_gen.config
dpezely: Removed with commit c300421

rebar.config
dpezely: Removed with commit 2181801

Confirmed: `make clean ; rebar3 grpc gen` still works

src/grpc/helium_packet_router_service.erl
macpie: ok I think we will be able to use this, this is just boilerplate to start including GRPCwe actually won't even use this service in there

rebar.config
```diff
@@ -1,5 +1,12 @@
+{cover_export_enabled, true}.
+
+{cover_enabled, true}.
+
 {deps, [
-    {lager, "3.9.2"}
+    {grpcbox, {git, "https://github.com/novalabsxyz/grpcbox.git"}},
+    {helium_proto, {git, "https://github.com/helium/proto.git"}},
+    {lager, "3.9.2"},
+    {lorawan, {git, "https://github.com/helium/erlang-lorawan.git"}}
```
macpie, line 9: I don't think we need this one in there I ll be adding it as part of another PR

config/sys.config
```diff
@@ -1,4 +1,30 @@
 [
+    {grpcbox, [
```
macpie, line 2: we should include this in `sys.config.src` as well as it will be used for release

dpezely: Thanks!  I keep forgetting about `sys.conf.src`...

Fixed in commit 3884fc8

rebar.config
dpezely: Removed dependency.  Fixed in commit 3884fc8

rebar.config
dpezely: (as above: let's merge with the current balance of Makefile vs rebar.config and revisit in a future PR if needed)

macpie (approved):

PR #15, opened 2022-07-15
make grpc before rebar compile

dpezely:
Enable gRPC build before `rebar3 compile` in GitHub CI test builds, and add grpc/protobufs things to `Makefile`, even though the protobufs artifacts get added in a different PR.

This is needed so PR#9 can build since it adds grpc things.  Apparently, GitHub CI changes must be in `main` branch in order to be used.  

However, it seems that we have a chicken vs egg situation (bootstrapping paradox) where GitHub CI won't fully use the new thing yet doesn't build properly without it.  Since **files in this PR are an excerpt** of [PR#9](https://github.com/helium/helium-packet-router/pull/9), maybe just merge that one and iterate as needed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,9 @@ jobs:
         with:
           toolchain: stable
       - name: build
-        run: ./rebar3 compile
+        run: |
+          make grpc
+          ./rebar3 compile
       - name: build test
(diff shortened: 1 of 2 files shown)


michaeldjeffrey (dismissed):

macpie:
Duplicate of https://github.com/helium/helium-packet-router/pull/9/files

Issue #16, opened 2022-07-21
Create Debian package for HPR


macpie:
See [PR#17](https://github.com/helium/helium-packet-router/pull/17)

Tasks:
- [x] construct Debian package `.deb` file based upon a successful `rebar3 release` run
   + Used `router` repo as placeholder: `rebar3 as testnet rel`;
   + See https://github.com/helium/router/pull/847
- [x] confirmed .deb generation against `router` repo; see Draft PR
- [x] authored `systemd` config and launch scripts (required)
- [x] authored `/etc/init/` and `/etc/rc?.d/` launch scripts
- [x] authored `Dockerfile` for testing within a pristine container
- [x] authored `Vagrantfile` for testing within a pristine VM
- [x] confirmed .deb extraction within clean Docker image
- [x] confirmed .deb extraction within clean Vagrant VM
- [ ] confirmed successful deploy of .deb within clean container
   + Blocked on Helium packet-router `main` branch having successful `make rel`
- [ ] confirmed successful deploy of .deb within clean VM
   + Blocked on Helium packet-router `main` branch having successful `make rel`

Future efforts:

- Local repo for storage until accepted into a Debian repo
- Successful submission to official Debian repo
   + For a rough idea of Debian package submission complexity: 
      https://www.debian.org/doc/manuals/packaging-tutorial/packaging-tutorial.en.pdf#Outline0.6.1.42

OUTDATED sketch:

Debian packages can be generated via `Makefile`, and here's a generic version that just needs to be confirmed against current `man` pages and refined:

```Makefile
# Excerpt for Makefile to build .deb package

# FIXME: don't use var in this way.  this is just to highlight it in this draft:
MAIN_THING=helium-packet-router

RELEASE=0.1.0
PACKAGE_BUILD=`date +%Y%m%d.%H%M%S`
REVISION=${RELEASE}-${PACKAGE_BUILD}
DEBIAN_PACKAGE_NAME=helium-packet-router

DESTDIR=/opt/hpr

RUNTIME_USER=nobody

package: package-bin package-cfg


# FIXME: change names of .sh scripts
release: ${MAIN_THING}
	rm -rf release
	mkdir -p release${DESTDIR}/{bin,var/{log,run}}
	cp -p ${MAIN_THING}{,.ini,.sh} release${DESTDIR}/bin/

# FIXME: update to accommodate systemd
package-bin: release
	mkdir -p release/DEBIAN
	find release -type d | xargs chmod 755
[...]

PR #17, opened 2022-07-28
Add Debian packaging via .mk file

dpezely:
Add Debian packaging:
- Ticket: [hpr#16](https://github.com/helium/helium-packet-router/issues/16)
- Keeps all of the Debian-ish procedures isolated from primary `Makefile`
- Copies file from `rebar3 release` into debian directory tree via `rsync` using hard-links to avoid duplicating files
- Usage: `make deb`

TO DO:
- [x] construct Debian package `.deb` file based upon a successful `rebar3 release` run
- [x] confirmed .deb generation against `router` repo; see below
- [ ] confirmed .deb installation on clean AWS instance
- [ ] ensure Debian-compliant version numbering
- [ ] `systemd` config and launch scripts (required)
- [ ] `/etc/init/` and `/etc/rc?.d/` launch scripts (optional, but good for community)
- [ ] Successful submission to official Debian repo (probably after merging this PR)

APPENDIX:

Extra bits used for making .deb for [router](https://github.com/helium/router) follow.

`../router/Make-deb-package.mk`

```Makefile
NAME=router
all:
	NAME=${NAME} SHORT_NAME=${NAME} \
	  make -f ../helium-packet-router/Make-deb-package.mk
```

`../router/priv/debian/deb-control.txt`
```conf
Package: helium-router
Version: 1.0-1
Section: lorawan
Priority: extra
Architecture: all
Depends: 
Suggests: esl-erlang
Installed-Size: 47185920
Maintainer: Nova Labs Inc <support@nova-labs.com>
Conflicts: helium-router
Provides: helium-router
Description: LoRaWAN Network Server (LNS)
 conforming to LoRaWAN 1.0.x specification.
 See https://github.com/helium/router
```

Run:
```bash
cd ~/helium/router/
make -f Make-deb-package.mk
```

Once this PR gets merged, we can back-port to `router` repo.

### DIFF
--- a/Make-deb-package.mk
+++ b/Make-deb-package.mk
@@ -1,5 +1,5 @@
 # Excerpt for Makefile to build .deb file https://wiki.debian.org/Packaging
-.PHONY: all package orig-tar prerelease release package-bin clean
+.PHONY: all deb package orig-tar prerelease release package-bin clean
 
 NAME ?= helium-packet-router
 SHORT_NAME ?= hpr
@@ -27,6 +27,8 @@ REBAR_VERSIONED_SCRIPT=${REBAR_RELEASE_DIR}/bin/${NAME}-${RELEASE}
 DEB_ARCH=$(shell dpkg --print-architecture)
 
 all: package
+
+deb: package
 
 package: orig-tar prerelease release package-bin
 
@@ -77,8 +79,9 @@ package-bin:
 	  >> release/DEBIAN/postinst
 	echo "ln -s ${DEST}/bin/${NAME}.sh /etc/rc1.d/K40${NAME}" \
 	  >> release/DEBIAN/postinst
-	echo "echo To manually start, use:" >> release/DEBIAN/postinst
-	echo "echo su - ${RUNTIME_USER} ${DEST}/bin/${NAME}.sh start" \
+	echo "echo To manually start, use:" \
+	  >> release/DEBIAN/postinst
+	echo "echo sudo -u ${RUNTIME_USER} ${DEST}/bin/${NAME}.sh foreground" \
 	  >> release/DEBIAN/postinst
 
 	echo "#! /bin/sh" > release/DEBIAN/prerm
@@ -88,6 +91,8 @@ package-bin:
 	echo "rm -f ${DEST}/var/log/${NAME}.log*" >> release/DEBIAN/prerm
 	echo "rm -f /etc/init.d/${NAME}.sh" >> release/DEBIAN/prerm
 	echo "rm -f /etc/rc?.d/*${NAME}" >> release/DEBIAN/prerm
+	echo 'userdel -r ${RUNTIME_USER}' >> release/DEBIAN/prerm
+
 	chmod 755 release/DEBIAN/p*
 
 	rsync -a --link-dest=$(shell pwd)/${REBAR_BUILD_DIR} \
--- a/Make-deb-package.mk
+++ b/Make-deb-package.mk
@@ -1,0 +1,101 @@
+# Excerpt for Makefile to build .deb file https://wiki.debian.org/Packaging
+.PHONY: all package orig-tar prerelease release package-bin clean
+
+NAME ?= helium-packet-router
+SHORT_NAME ?= hpr
+SNAKE_NAME=$(shell echo ${NAME} | sed 's/-/_/g')
+
+CURRENT_YEAR ?= $(shell date +"%Y")
+RELEASE ?= $(shell git tag -l ${CURRENT_YEAR}'*' | sort | tail -1)
+
+# ${DEST} filepath will be created beneath ./release subdirectory.
+DEST ?= /opt/${SHORT_NAME}
+RUNTIME_USER ?= helium
+
+# Before upload to Debian repo, manually change date to next sequence number:
+PACKAGE_BUILD=`date +%Y%m%d.%H%M%S`
+REVISION=${RELEASE}-${PACKAGE_BUILD}
+DEBIAN_PACKAGE_NAME=${NAME}
+ORIG_SRC_TAR="${NAME}_${RELEASE}.orig.tar.gz"
+
+# `rebar` is Erlang's build system.
+REBAR_BUILD_DIR=_build/default/rel
+REBAR_RELEASE_DIR=${REBAR_BUILD_DIR}/${SNAKE_NAME}
+REBAR_LAUNCH_SCRIPT=${REBAR_RELEASE_DIR}/bin/${NAME}
+REBAR_VERSIONED_SCRIPT=${REBAR_RELEASE_DIR}/bin/${NAME}-${RELEASE}
+
+DEB_ARCH=$(shell dpkg --print-architecture)
+
+all: package
+
+package: orig-tar prerelease release package-bin
+
+# Preliminary to the packaging workflow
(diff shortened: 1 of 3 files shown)


Make-deb-package.mk
```diff
@@ -0,0 +1,106 @@
...
+	echo "${DEST}/bin/${NAME}.sh unkind-stop" >> release/DEBIAN/prerm
+	echo "${DEST}/bin/${NAME}.sh clean-logs" >> release/DEBIAN/prerm
+	echo "rm -f ${DEST}/var/log/${NAME}.log*" >> release/DEBIAN/prerm
+	echo "rm -f /etc/init.d/${NAME}.sh" >> release/DEBIAN/prerm
+	echo "rm -f /etc/rc?.d/*${NAME}" >> release/DEBIAN/prerm
+	echo 'userdel -r ${RUNTIME_USER}' >> release/DEBIAN/prerm
+
+	chmod 755 release/DEBIAN/p*
+
+	rsync -a --link-dest=$(shell pwd)/${REBAR_BUILD_DIR} \
```
jontow, line 98: very optional idea: rsync being common but not always preinstalled, maybe we could check for its presence in the `prerelease` target and bail if not found, before spending time compiling anything? envisioning something like this:  `rsync --help >/dev/null 2>&1 || echo "rsync not installed or missing from PATH"`

Make-deb-package.mk
```diff
@@ -0,0 +1,106 @@
...
+	  ${REBAR_BUILD_DIR}/ release${DEST}/
+
+	dpkg-deb --build release
```
jontow, line 101: According to the manpage, I think 101 and 102 can be combined here to avoid the mv(1):
`dpkg-deb --build release ${DEBIAN_PACKAGE_NAME}_${REVISION}_${DEB_ARCH}.deb`

jontow (dismissed):
way nice first pass, added some possible improvement comments, but nothing that stops my :+1:

Make-deb-package.mk
dpezely: Nice!  Fixed in commit a544c77

Make-deb-package.mk
dpezely: I went with `which` instead of `--help` which should accomplish the same thing.

Fixed in commit a544c77

Make-deb-package.mk
dpezely: And that fix was corrected with commit f93c870

Make-deb-package.mk
dpezely: And that fix was corrected with commit f93c870

dpezely:
This effort is **blocked** until Helium packet-router "main" branch can run `make rel` successfully.

dpezely:
> This effort is **blocked** until Helium packet-router "main" branch can run `make rel` successfully.

@jadeallenx Rumor has it that you're versed in `relx`.  Would you be able to help drag this over the finish line for making a simple Debian package, please?

TL;DR: 
- Checklist in ticket is up-to-date: https://github.com/helium/helium-packet-router/issues/16
- Run `make deb` or for more nuanced commands, use: `make -f Make-deb-package.mk vm-test vm-ssh vm-relx`
- Background:
  + I thought `relx` bits were successful based upon `router background` *apparently* returning cleanly but silently exiting with `127`
  + Based upon that perception of the (false) positive launch, I wrote all the other things: systemd service, init.d & rc*.d scripts, Vagrantfile, etc.
  + `rebar3 relx` successfully creates something runnable within Docker
  + However, using `make -f Make-deb-package.mk vm-relx` which is a simple Vagrant VM wrapped around that same `rel` directory tree, nothing runs / always exits with status `127`.
  + Likewise, the resulting `.deb` file fails to launch the app during install

dpezely (commented):
Excellent!  Thank you!

Since I created the initial PR, GitHub won't allow me to approve despite Jade overhauling and finishing it.

.github/scripts/make_deb.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+VERSION=$(git describe)
+
+DIAGNOSTIC=1 ./rebar3 release
```
michaeldjeffrey, line 7: Do we need to use `make rel` here to make sure grpc deps are built?

deb/after_install.sh
```diff
@@ -0,0 +1,2 @@
+# add miner to /usr/local/bin it appears in path
```
michaeldjeffrey, line 1:
```suggestion
# add hpr to /usr/local/bin it appears in path
```

deb/before_upgrade.sh
```diff
@@ -0,0 +1,2 @@
+# add system user for file ownership and systemd user, if not exists
+useradd --system --home-dir /opt/miner --create-home helium || true
```
michaeldjeffrey, line 2:
```suggestion
useradd --system --home-dir /opt/hpr --create-home helium || true
```

deb/after_upgrade.sh
```diff
@@ -0,0 +1,2 @@
+# add miner to /usr/local/bin so it appears in path, if it does not already exist
```
michaeldjeffrey, line 1:
```suggestion
# add hpr to /usr/local/bin so it appears in path, if it does not already exist
```

dpezely (commented):
There's `config/sys.config.src`, but does this also need `config/sys.config` as well?

(I'm still baffled about why there would be two such files-- why, Erlang, why?)

rebar.config
```diff
@@ -3,6 +3,7 @@
 {cover_enabled, true}.
 
 {deps, [
+    recon,
```
michaeldjeffrey, line 6:
```suggestion
```

We already have recon pinned to a version at the bottom of the deps.

jadeallenx:
> (I'm still baffled about why there would be two such files-- why, Erlang, why?)

A `.src` file is a template which rebar uses to create the "real" sys.config at runtime - it is not natively supported by ERTS; the `sys.config` file is fhe old classic file which relx still supports. You need one or the other, not both, although there are plenty of projects which have both. `relx` prefers `.src` though and if it exists, will use it instead of plain ol' `sys.config.

michaeldjeffrey (dismissed):

config/sys.config.src
```diff
@@ -1,6 +1,6 @@
 [
     {hpr, [
-        {base_dir, "/var/data"}
+        {base_dir, "data"}
```
macpie, line 3: base_dir os not needed anymore

rebar.config
```diff
@@ -67,12 +67,12 @@
...
 
 {relx, [
-    {release, {hpr, "1.0.0"}, [hpr]},
+    {release, {hpr, git}, [hpr, runtime_tools, tools, recon]},
     {vm_args, "config/vm.args"},
-    {sys_config, "config/sys.config"},
     {sys_config_src, "config/sys.config.src"},
-    {include_src, true},
-    {overlay, []},
+    {include_src, false},
```
macpie, line 73: Does this mean we won't be able to hot load?

jadeallenx: Don't think so, just makes the image size as small as possible - by hot loading do you mean using sasl and systools like `relujp` or just sideloading new beams?

macpie: > sideloading new beams

macpie (approved):

PR #19, opened 2022-08-01
Add Loki support

macpie:


config/sys.config
```diff
@@ -1 +1,37 @@
...
+                {file, "info.log"},
+                {level, info},
+                {size, 0},
+                {date, "$D0"},
+                {count, 7},
+                {formatter_config, [
+                    "time=",
+                    date,
+                    "T",
+                    time,
```
michaeldjeffrey, line 17: We can replace a few of these with a the atom `metadata` from lager.
ref: https://github.com/erlang-lager/lager/blob/master/src/lager_default_formatter.erl#L96-L100

PR #22, opened 2022-08-03
Routing Config based on Proto

### LINKED ISSUE #6: Routing Config based on Proto
- [ ] Create proto route matching 
```
{
        net_id=3,
        dev_ranges=[{12, 20}, ..]
        euis=[{appeui, deveui}, ...] 
        lns_address="lora.com|10.0.0.9"
        protol="gwmp|http"
        oui=1
}
```
- [ ] Save config locally

macpie:
Fix #6


mikev (dismissed):

src/hpr_routing_config.erl
```diff
@@ -0,0 +1,257 @@
...
+init_ets() ->
+    {EUIRoutes, DevAddrRoutes} = dets:foldl(
+        fun({_OUI, Route}, {EUIRoutesAcc, DevAddrRoutesAcc}) ->
+            EUIs = hpr_routing_config_route:euis(Route),
+            EUIRoutes = [{{AppEUI, DevEUI}, Route} || {AppEUI, DevEUI} <- EUIs],
+            NetID = hpr_routing_config_route:net_id(Route),
+            DevAddrRangeRoutes =
+                case hpr_routing_config_route:devaddr_ranges(Route) of
+                    [] ->
+                        [{{NetID, 0, 0}, Route}];
```
michaeldjeffrey, line 151: When there's no range, we can insert min and max values.
ref: https://github.com/helium/packet-purchaser/blob/master/src/pp_config_v2.erl#L108

This allows the spec when looking up DevAddr ranges to only need to do 2 comparisons instead of 4. 

When we look up DevAddrs by range, we already have the NetID, so the possibility of sending data to the wrong range is very low.

barndon (dismissed):

src/hpr_routing.erl
```diff
@@ -0,0 +1,112 @@
...
+        true ->
+            execute_checks(Packet, Rest)
+    end.
+
+-spec packet_type(Packet :: hpr_packet_up:packet()) ->
+    {join_req, non_neg_integer(), non_neg_integer()} | {uplink, non_neg_integer()} | undefined.
+packet_type(Packet) ->
+    case hpr_packet_up:payload(Packet) of
+        <<?JOIN_REQUEST:3, _:5, AppEUI:64/integer-unsigned-little,
+            DevEUI:64/integer-unsigned-little, _DevNonce:2/binary, _MIC:4/binary>> ->
```
dpezely, line 89: @mikev Which library call from `erlang-lorawan` library would you recommend be used here instead?

(We should use destructuring-bind via native Erlang binary as last resort.)

dpezely (approved):
My focus on this code was from an overall architecture perspective: recognized patterns from `router` and liking the naming conventions such as `LNS` / `lns` in keeping with LoRaWAN Spec terminology-- very nice!

Tests will likely evolve with new feature code, so I didn't spend much quality time on those on this first pass.

dpezely (dismissed):

dpezely (approved):

dpezely (approved):

PR #24, opened 2022-08-11
hwahby/packet reporting

hwahby:
WIP: Packet reporter will periodically write summary data of each packet routed through Packet Router to S3 for use by Rewards Service.


dpezely (approved):
Yes, this is still a Draft PR but wanted to let you know that someone looked it over...

config/sys.config.src
```diff
@@ -1,5 +1,10 @@
 [
     {hpr, [
+        {aws_config, [
+            {access_key_id, <<"${AWS_ACCESS_KEY_ID}">>},
```
macpie, line 4: Do you want binaries here? It seems like in other config you are using strings

macpie:
so far so good, just missing the proto stuff now?

config/sys.config.src
hwahby: Yes, they should be binaries. Fixed to make types consistent.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,302 @@
...
+    file:delete(FilePath),
+    {noreply, State};
+handle_cast(_Request, State = #state{}) ->
+    {noreply, State}.
+
+handle_info(_Info, State = #state{}) ->
+    {noreply, State}.
+
+terminate(_Reason, _State = #state{}) ->
+    ok.
```
michaeldjeffrey, line 176: We may want to look at the reason here and try to write the last of what's in ETS to file before we shut down and lose it.

dpezely (commented):
The `.env` file should be in `.gitignore` (and excluded from PR) to avoid accidentally committing one with real keys.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+
+    case FileSize >= MaxFileSize of
+        true -> ?MODULE:upload_packets(FilePath);
+        false -> skip
+    end,
+    {noreply, State};
+handle_cast({upload, FilePath}, State = #state{aws_client = AWSClient}) ->
+    Timestamp = erlang:system_time(millisecond),
+    FileName = list_to_binary("packetreport." ++ integer_to_list(Timestamp) ++ ".gz"),
+    upload_file(AWSClient, FilePath, FileName),
```
michaeldjeffrey, line 106: We shouldn't delete the file if we can't upload it. 
Maybe a TODO here for some retry logic similar to what they're doing in rust.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+    FilePath = "./data/tmp/packetreport." ++ integer_to_list(Timestamp) ++ ".txt",
+
+    Packet = test_utils:join_packet_up(#{}),
+    Packet2 = test_utils:uplink_packet_up(#{}),
+    Route = test_utils:packet_route(#{}),
+
+    report_packet(Packet, Route),
+    report_packet(Packet2, Route),
+
+    timer:sleep(1000),
```
michaeldjeffrey, line 287: Do we need these sleeps? `report_packet/2` is a sync call to ets.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+
+    handle_write(FilePath, []),
+
+    {ok, Content} = file:read_file(FilePath),
+
+    2 = length(binary:split(Content, <<"\n">>, [trim, global])),
```
michaeldjeffrey, line 293: Can we put this and L307 in `?assertEqual`.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+
+    {ok, Content2} = file:read_file(FilePath),
+
+    4 = length(binary:split(Content2, <<"\n">>, [trim, global])),
+
+    file:delete(FilePath),
+
+    ok.
+
+% aws_test() ->
```
michaeldjeffrey, line 313: If this is for testing with the localstack setup, can we some documentation around that?
If it's not, and unused, this can be removed.

test/test_utils.erl
```diff
@@ -75,6 +77,18 @@ uplink_packet_up(Opts0) ->
     SigFun = maps:get(sig_fun, Opts0, fun(_) -> <<"signature">> end),
     hpr_packet_up:sign(PacketUp, SigFun).
 
+-spec packet_route(Opts :: map()) -> hpr_packet_route:route().
+packet_route(Opts0) ->
+    NetID = maps:get(net_id, Opts0, 0),
+    DevAddrRanges = maps:get(devaddr_ranges, Opts0, [
+        #packet_router_route_devaddr_range_v1_pb{start = 1, 'end' = 10}
+    ]),
+    EUIs = maps:get(euis, Opts0, [#packet_router_route_eui_v1_pb{app_eui = 1, dev_eui = 1}]),
```
michaeldjeffrey, line 86:
```suggestion
    DevAddrRanges = maps:get(devaddr_ranges, Opts0, [{1, 10}]),
    EUIs = maps:get(euis, Opts0, [{1, 1}]),
```
It wasn't breaking because the list comprehension was discarding in `hpr_route:new/6` was bypassing the values that didn't match. Defaults here would always result in empty EUI and DevAddr lists.

src/hpr_packet_reporter.erl
ruttenberg: @michaeldjeffrey 
What's the advantage of ?assertEqual ?

michaeldjeffrey: It helps mark these checks as involved in the unit under test rather than sanity checks that are hanging around.
And gives some nicer feedback if they fail. Rather than a `badmatch` we get the values with the context that will hopefully explain why we care that these values don't match.

src/hpr_sup.erl
```diff
@@ -33,6 +31,23 @@
...
 }).
 
+-define(WORKER(I, Mod, Args), #{
+    id => I,
+    start => {Mod, start_link, Args},
+    restart => permanent,
+    shutdown => 5000,
+    type => worker,
+    modules => [I]
+}).
```
ruttenberg, line 41: The arity 3 version of ?WORKER is not used.
It looks like it should be deleted.

src/hpr_routing.erl
```diff
@@ -71,6 +71,7 @@ handle_packet(Packet, StreamHandler) ->
 deliver_packet(_Packet, _StreamHandler, []) ->
     ok;
 deliver_packet(Packet, StreamHandler, [Route | Routes]) ->
+    hpr_packet_reporter:report_packet(Packet, Route),
```
ruttenberg, line 74: @michaeldjeffrey 
I can take care of the FIXME on L85 now, unless there is something more pressing for me to do.

src/hpr_packet_reporter.erl
ruttenberg: Makes sense. Thanks.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+
+-spec init_ets() -> ok.
+init_ets() ->
+    ?ETS = ets:new(?ETS, [named_table, bag, public, {write_concurrency, true}]),
+    ok.
+
+-spec report_packet(hpr_packet_up:packet(), hpr_route:route()) -> ok.
+report_packet(Packet, Route) ->
+    EncodedPacket = encode_packet(Packet, Route),
+    Timestamp = erlang:system_time(millisecond),
```
ruttenberg, line 55: @hwahby 
It looks like you could get the same Timestamp from different calls to _erlang:system_time_.
That would make this Timestamp a bad choice for a key.

ruttenberg: Oops, sorry. I misread the docs for 'bag'.
Comment withdrawn!

src/hpr_routing.erl
michaeldjeffrey: Go for it!!

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,331 @@
...
+
+% ------------------------------------------------------------------
+% EUNIT Tests
+% ------------------------------------------------------------------
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("common_test/include/ct.hrl").
+
+file_test() ->
+    ?ETS = ets:new(?ETS, [named_table, bag, public, {write_concurrency, true}]),
```
ruttenberg, line 276: This test fails.
``` Failure/Error: {error,
                        {badmatch,{error,lager_not_running}},
                        [{hpr_packet_reporter,handle_write,2,```

File: src/hpr_routing.erl (hunk 9 above)
Comment (ruttenberg) [reply src/hpr_routing.erl:74]: This has been done in
https://github.com/helium/helium-packet-router/pull/46

File: src/hpr_packet_reporter.erl [hunk 12]
```diff
@@ -0,0 +1,328 @@
...
+stop_report_interval() ->
+    gen_server:cast(?SERVER, stop_interval).
+
+%%%===================================================================
+%%% Spawning and gen_server implementation
+%%%===================================================================
+
+init(#{base_dir := BaseDir} = _Args) ->
+    AWSClient = setup_aws(),
+    WriteDir = filename:join(BaseDir, "tmp/"),
```
ruttenberg, line 80: The trailing slash in "tmp/" will be removed by filename:join/2, so it would be clearer to make the second argument  "tmp"

src/hpr_sup.erl
macpie: 👍🏻

src/hpr_packet_reporter.erl
macpie: 👍🏻

macpie:
I think this could use a end to end CT, using the `localstack`

config/sys.config
```diff
@@ -1,5 +1,14 @@
...
     {hpr, [
+        {aws_config, [
+            {access_key_id, <<"testkey">>},
+            {secret_access_key, <<"testsecret">>},
+            {aws_region, <<"us-east-1">>}
+        ]},
+        {packet_reporter_bucket, <<"bucket-name">>},
+        {packet_reporter_report_interval, 300000},
+        {packet_reporter_max_file_size, 50_000_000},
+        {packet_reporter_upload_window, 900000},
```
macpie, line 11:
```suggestion
    {packet_reporter, #{
        access_key_id => <<"testkey">>,
        secret_access_key => <<"testsecret">>,
        region => <<"us-east-1">>,
        bucket =>  <<"bucket-name">>,
        report_interval => 300000,
        max_file_size => 50_000_000,
        upload_window => 900000
    }}
```

test/hpr_packet_reporter_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+    end_per_testcase/2
+]).
+
+-export([
+    upload_report_test/1,
+    upload_window_test/1
+]).
+
+-include("hpr.hrl").
+-include("../src/grpc/autogen/server/packet_router_pb.hrl").
```
macpie, line 15: I want to try to avoid using records here and instead use the `her_router` module

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+-spec handle_restart_interval(
+    IntervalRef :: timer:tref() | undefined, IntervalDuration :: interval_duration_ms()
+) -> {ok, timer:tref()}.
+handle_restart_interval(undefined, IntervalDuration) ->
+    start_report_interval(IntervalDuration);
+handle_restart_interval(IntervalRef, IntervalDuration) ->
+    timer:cancel(IntervalRef),
+    start_report_interval(IntervalDuration).
+
+-spec handle_stop_interval(IntervalRef :: timer:tref()) -> {ok, cancel} | {error, term()}.
```
macpie, line 270: Spec is off I think

docker-compose-local.yaml
```diff
@@ -0,0 +1,27 @@
+version: '3'
```
macpie, line 1: I don't think we need this compose

test/test_utils.erl
```diff
@@ -5,10 +5,13 @@
     end_per_testcase/2,
     join_packet_up/1,
     uplink_packet_up/1,
+    packet_route/1,
     match_map/2
 ]).
 
 -include("hpr.hrl").
+-include("../src/grpc/autogen/server/packet_router_pb.hrl").
```
macpie, line 13: same as above

test/hpr_packet_reporter_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+    ets:delete_all_objects(hpr_packet_report_ets),
+    meck:unload(aws_s3),
+    meck:unload(hpr_packet_reporter),
+
+    test_utils:end_per_testcase(TestCase, Config).
+
+%%--------------------------------------------------------------------
+%% TEST CASES
+%%--------------------------------------------------------------------
+
```
macpie, line 72: Could we have some kind of negative test is AWS down for ex

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+        report_interval = ReportInterval,
+        max_file_size = MaxFileSize
+    }
+) ->
+    handle_write(FilePath, MaxFileSize, ?FILE_WRITE_OPTIONS),
+    handle_stop_interval(ReportInterval),
+    lager:warning("packet reporter process terminated: ~s", [Reason]),
+    ok.
+
+code_change(_OldVsn, State = #state{}, _Extra) ->
```
marcsugiyama, line 196: `code_change/3` callback is optional. You can remove this.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+-spec report_packet(Packet :: hpr_packet_up:packet(), PacketRoute :: hpr_route:route()) -> ok.
+report_packet(Packet, PacketRoute) ->
+    EncodedPacket = encode_packet(Packet, PacketRoute),
+    ReportTimestamp = erlang:system_time(millisecond),
+    true = ets:insert(?ETS, {ReportTimestamp, EncodedPacket}),
+    ok.
+
+-spec write_packets() -> ok.
+write_packets() ->
+    gen_server:cast(?SERVER, write).
```
marcsugiyama, line 94: Style: should the message match the API function name? It makes it a little easier to find the matching callback.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+    FileName = list_to_binary("packetreport." ++ integer_to_list(UploadTimestamp) ++ ".gz"),
+
+    case upload_file(AWSClient, list_to_binary(FilePath), FileName, Bucket) of
+        {ok, _} ->
+            file:delete(FilePath);
+        Error ->
+            lager:warning("packet reporter failed to upload: ~p~n", [Error]),
+            error
+    end,
+    {noreply, State#state{upload_window_start_time = UploadTimestamp}};
```
marcsugiyama, line 170: If the upload fails, should we still advance the upload window start time? I'm thinking we should not so the next time we check we'll try again, rather than waiting an entire interval.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+            gateway_timestamp_ms = GatewayTimestamp,
+            oui = OUI,
+            net_id = NetID,
+            rssi = RSSI,
+            frequency = FrequencyMhz,
+            datarate = Datarate,
+            snr = SNR,
+            region = Region,
+            gateway = Gateway,
+            payload_hash = crypto:hash(sha256, Payload)
```
michaeldjeffrey, line 302:
```suggestion
            payload_hash = hpr_packet_up:phash(Packet)
```

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+handle_stop_interval(undefined) -> {ok, no_interval};
+handle_stop_interval(IntervalRef) -> timer:cancel(IntervalRef).
+
+-spec encode_packet(Packet :: hpr_packet_up:packet(), PacketRoute :: hpr_route:route()) -> binary().
+encode_packet(
+    #packet_router_packet_up_v1_pb{
+        payload = Payload,
+        timestamp = GatewayTimestamp,
+        rssi = RSSI,
+        frequency = FrequencyMhz,
```
michaeldjeffrey, line 280: Frequency is in Hz.
For both `packet_up` and `packet_report`.

src/hpr_routing.erl
```diff
@@ -88,16 +88,25 @@ deliver_packet(Packet, [Route | Routes]) ->
         "delivering packet to ~s",
         [hpr_route:lns(Route)]
     ),
-    %% FIXME: delivery could be halted to multiple routes if one of the earlier
-    %% Protocol:send(...) errors out.
+    hpr_packet_reporter:report_packet(Packet, Route),
```
michaeldjeffrey, line 91: We're going to need to move this to make sure we don't account for packets that failed to be delivered.
@macpie I think you were thinking about where it could go?

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+%%%===================================================================
+%%% Internal Function Definitions
+%%%===================================================================
+
+-spec handle_write(FilePath :: string(), MaxFileSize :: integer(), WriteOptions :: [atom()]) ->
+    {ok, FileSize :: integer()}.
+
+handle_write(FilePath, MaxFileSize, WriteOptions) ->
+    WriteTimestamp = erlang:system_time(millisecond),
+    {ok, S} = open_tmp_file(FilePath, WriteOptions),
```
marcsugiyama, line 208: we'll crash here if the temp file can't be opened. Should we be more defensive here and look for the error tuple?

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+            snr = SNR,
+            region = Region,
+            gateway = Gateway,
+            payload_hash = crypto:hash(sha256, Payload)
+        },
+        packet_router_packet_report_v1_pb
+    ).
+
+-spec encode_packet_size(EncodedPacket :: binary()) -> PacketSize :: binary().
+encode_packet_size(EncodedPacket) ->
```
marcsugiyama, line 308: Is this the same as `<<PacketSize:32/big-integer-unsigned>>`?

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+handle_write(FilePath, MaxFileSize, WriteOptions) ->
+    WriteTimestamp = erlang:system_time(millisecond),
+    {ok, S} = open_tmp_file(FilePath, WriteOptions),
+    Data = get_packets_by_timestamp(WriteTimestamp),
+
+    lists:foreach(
+        fun(Packet) ->
+            PacketSize = encode_packet_size(Packet),
+            file:write(S, PacketSize),
+            file:write(S, Packet)
```
marcsugiyama, line 215: I think you can combine these into a single call: `file:write(S, [PacketSize, Packet])`. The data for `file:write` is `iodata()`.

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+
+-spec upload_window_elapsed(StartTime :: timestamp_ms(), UploadWindow :: timestamp_ms()) ->
+    WindowElapsed :: boolean().
+upload_window_elapsed(StartTime, UploadWindow) ->
+    Timestamp = erlang:system_time(millisecond),
+    Timestamp - StartTime >= UploadWindow.
+
+-spec get_packets_by_timestamp(Timestamp :: timestamp_ms()) -> [term()].
+get_packets_by_timestamp(Timestamp) ->
+    ets:select(?ETS, [{{'$1', '$2'}, [{'=<', '$1', Timestamp}], ['$2']}]).
```
marcsugiyama, line 370: Since the table is a `bag`, I suspect that the rows will not be returned in timestamp order. Is that acceptable? If not, you'd have to figure out a way of making the key values unique and use `ordered_set` or sort them after the `ets:select`.

marcsugiyama: Do we expect this `select` to return a huge number of rows? Should we limit the return count with `select/3` and then iterate on the continuation?

src/hpr_packet_reporter.erl
```diff
@@ -0,0 +1,432 @@
...
+
+% ------------------------------------------------------------------
+% EUNIT Tests
+% ------------------------------------------------------------------
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("common_test/include/ct.hrl").
+
+file_test() ->
+    ?ETS = ets:new(?ETS, [named_table, bag, public, {write_concurrency, true}]),
```
marcsugiyama, line 384: Use `init_ets/0`?

test/hpr_packet_reporter_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+    meck:new(hpr_packet_reporter, [passthrough]),
+
+    Config1.
+
+%%--------------------------------------------------------------------
+%% TEST CASE TEARDOWN
+%%--------------------------------------------------------------------
+end_per_testcase(TestCase, Config) ->
+    ets:delete_all_objects(hpr_packet_report_ets),
+    meck:unload(aws_s3),
```
marcsugiyama, line 64: `meck:unload/0` unloads all mocked modules. This might be safer in case a test ends up mocking a module.

test/hpr_packet_reporter_SUITE.erl
```diff
@@ -0,0 +1,187 @@
...
+        fun({_, P}) ->
+            P
+        end,
+        ETSValues
+    ),
+    verify_packet(Packet, Route, P1),
+    verify_packet(Packet2, Route, P2),
+
+    %% Encoded packets are written to a tmp write file
+    hpr_packet_reporter:handle_cast(write, State),
```
marcsugiyama, line 97: Use the public API instead? `hpr_packet_report:write_packets/0`? Same for upload?

marcsugiyama: Ah right, they are casts. I hadn't considered that. There's no easy way to deal with that except to add an explicit yield or short sleep. Given that. I think it's fine to call handle_cast directly here.

michaeldjeffrey: You could do a test only export of the writing and uploading functions. And directly call those.

src/hpr_packet_reporter.erl
hwahby: Very cool, changing.

test/hpr_packet_reporter_SUITE.erl
hwahby: Sure, will add one.

src/hpr_packet_reporter.erl
hwahby: Good points, taking a look at my ETS usage.

src/hpr_packet_reporter.erl
hwahby: This could be an option. 
Upload failures are retried up to five times with a minute between retries. Per conversations with Madninja, the packets may be sufficiently stale after that point that reporting them wouldn't be a priority.

When upload is called the write file is rotated, so calling upload on the next interval (as is) would upload the new write file. I could see incorporating some kind of retry queue as an improvement.

src/hpr_routing.erl
macpie: Yes we should move this to line 114 when we got an "ok" back from the protocol

src/hpr_packet_reporter.erl
macpie: yes probably

config/sys.config.src
```diff
@@ -1,9 +1,21 @@
...
+            max_file_size => 50_000_000,
+            upload_window => 900000,
+            retry_sleep_time => 60000,
+            max_upload_retries => 5
+        }},
         {config_worker, #{
             host => "${HPR_CONFIG_WORKER_HOST}",
             port => "${HPR_CONFIG_WORKER_PORT:-8080}",
-            file_backup_path => "/var/data/config_worker.backup"
+            file_backup_path => "/var/data/hpr/config_worker.backup"
```
macpie, line 18:
```suggestion
            file_backup_path => "/var/data/config_worker.backup"
```
This needs to be removed or it will mess things up in container

config/sys.config.src
```diff
@@ -1,5 +1,17 @@
 [
     {hpr, [
+        {packet_reporter, #{
+            access_key_id => <<"${AWS_ACCESS_KEY_ID}">>,
+            secret_access_key => <<"${AWS_SECRET_ACCESS_KEY}">>,
+            region => <<"${AWS_DEFAULT_REGION}">>,
+            bucket => <<"${PACKET_REPORTER_BUCKET_NAME}">>,
+            interval_duration => 300000,
+            write_dir => "/var/data/hpr/tmp/",
```
macpie, line 9:
```suggestion
            write_dir => "/var/data/tmp/",
```

config/ct.config
```diff
@@ -1,4 +1,20 @@
 [
+    {hpr, [
+        {packet_reporter, #{
+            access_key_id => <<"testkey">>,
+            secret_access_key => <<"testsecret">>,
+            region => <<"local">>,
+            bucket => <<"test-bucket">>,
+            interval_duration => 300000,
+            write_dir => "/var/data/hpr/tmp/",
```
macpie, line 9:
```suggestion
            write_dir => "/var/data/tmp/",
```

macpie:
For ref https://github.com/novalabsxyz/ops/issues/258#issuecomment-1245688548

src/hpr_packet_reporter.erl
hwahby: Made the suggested changes:
1. Switched to an `ordered_set`, incrementing a counter to act as index in ETS.
2. Adjusted select patterns to use `<` to avoid the possible race condition.
3. Rewrote to write to the file in batches using `select/3`

michaeldjeffrey (approved):

PR #25, opened 2022-08-12
Ruttenberg/helium packet router issue 004

ruttenberg:


michaeldjeffrey:
We been discussing the api of udp_worker, and have decided for now on:

```erlang
hpr_udp_worker:push_data(
              UDPWorkerPid,
              packet_up_to_rxpk(PacketUp),  % data will be given to worker in sendable format
              HandlerPid,
              {"127.0.0.1", 1700}
)
```

src/protocols/gwmp/hpr_gwmp_worker.erl
```diff
@@ -0,0 +1,500 @@
...
+-define(SERVER, ?MODULE).
+
+-define(METRICS_PREFIX, "helium_packet_router_").
+
+-type pull_data_map() :: #{
+    gwmp_udp_socket:socket_dest() => #{timer_ref := reference(), token := binary()}
+}.
+
+-record(state, {
+    location :: no_location | {pos_integer(), float(), float()} | undefined,
```
michaeldjeffrey, line 44: We can remove this from the worker for now.

src/protocols/gwmp/hpr_gwmp_worker.erl
```diff
@@ -0,0 +1,500 @@
...
+    pubkeybin :: libp2p_crypto:pubkey_bin(),
+    socket :: gwmp_udp_socket:socket(),
+    push_data = #{} :: #{binary() => {binary(), reference()}},
+    response_handler_pid :: undefined | pid(),
+    pull_resp_fun :: undefined | function(),
```
michaeldjeffrey, line 49: Make this a direct call now that we're no longer pulling this worker from the utils repo.

src/protocols/gwmp/hpr_gwmp_worker.erl
```diff
@@ -0,0 +1,500 @@
...
+
+-spec schedule_pull_data(non_neg_integer(), gwmp_udp_socket:socket_dest()) -> reference().
+schedule_pull_data(PullDataTimer, SocketDest) ->
+    _ = erlang:send_after(PullDataTimer, self(), {?PULL_DATA_TICK, SocketDest}).
+
+-spec schedule_shutdown(non_neg_integer()) -> reference().
+schedule_shutdown(ShutdownTimer) ->
+    _ = erlang:send_after(ShutdownTimer, self(), ?SHUTDOWN_TICK).
+
+update_address(Socket0, Address, Port) ->
```
michaeldjeffrey, line 327: I'm wondering if we still need the `gwmp_udp_socket`. 
The only feature we're really using from it is keeping the location with the socket, but we're back in a position where we have to update the address before sending anything. 

It might be nice to keep the module for the types it defines, but removing `update_address/2` and requiring the dest be supplied with each `send/3`.

src/protocols/hpr_gwmp_router.erl
```diff
@@ -0,0 +1,112 @@
...
+    Packet :: hpr_packet_up:packet(),
+    Stream :: grpcbox_stream:t(),
+    Routes :: hpr_route:route()
+) -> ok | {error, any()}.
+send(PacketUp, Stream, Route) ->
+    Hotspot = hpr_packet_up:hotspot(PacketUp),
+
+    case hpr_gwmp_udp_sup:maybe_start_worker(Hotspot, #{socket_dest => route_to_dest(Route)}) of
+        {error, Reason} ->
+            throw({gwmp_sup_err, Reason});
```
michaeldjeffrey, line 23: What can we reasonably do here?
I think throwing will crash _all_ the senders, and we don't want that.

src/protocols/hpr_gwmp_router.erl
```diff
@@ -0,0 +1,112 @@
...
+
+-spec send(
+    Packet :: hpr_packet_up:packet(),
+    Stream :: grpcbox_stream:t(),
+    Routes :: hpr_route:route()
+) -> ok | {error, any()}.
+send(PacketUp, Stream, Route) ->
+    Hotspot = hpr_packet_up:hotspot(PacketUp),
+
+    case hpr_gwmp_udp_sup:maybe_start_worker(Hotspot, #{socket_dest => route_to_dest(Route)}) of
```
michaeldjeffrey, line 21: Should `route_to_dest/1`  live in the udp_worker?
The worker knows what it needs to start.
And moving the args creation and use closer together is 1 less thing that could become mismatched in the future.

michaeldjeffrey (commented):
Pulling some tests from packet-purchaser would be nice. 
- Some tests that ensure we're turning `PacketRouterPacketUpV1` into `RxPk` properly, and that will break if we change those functions.
- Same for `TxPk` to `PacketRouterPacketDownV1` (they'll mostly likely break soon, but I want that).
- Other tests that feel nice and applicable.

src/protocols/gwmp/hpr_gwmp_worker.erl
ruttenberg: Done.

src/protocols/gwmp/hpr_gwmp_worker.erl
ruttenberg: Done.

src/protocols/gwmp/hpr_gwmp_worker.erl
```diff
@@ -0,0 +1,494 @@
...
+
+-type pull_data_map() :: #{
+    gwmp_udp_socket:socket_dest() => #{timer_ref := reference(), token := binary()}
+}.
+
+-record(state, {
+    pubkeybin :: libp2p_crypto:pubkey_bin(),
+    socket :: gwmp_udp_socket:socket(),
+    push_data = #{} :: #{binary() => {binary(), reference()}},
+    response_stream :: undefined | tuple(),
```
ruttenberg, line 47: @michaeldjeffrey 
I changed the name and type of this stream handler because it can't be a pid. It is passed as the third argument of grpcbox_stream:send/3.
Once the call was no longer wrapped in a closure, dialyzer started complaining.

src/protocols/hpr_gwmp_router.erl
ruttenberg: It doesn't feel right to have 
```    Lns = hpr_route:lns(Route),
```
in the worker. The udp worker shouldn't know about routes.

I'm not sure if this is an improvement, but you could pass the 'Lns' above to the worker, and then the worker can process it into the desired form of {Address, Port}.

src/protocols/hpr_gwmp_router.erl
ruttenberg: `hpr_routing:handle_packet/2` can return {error, any()}.
If `hpr_routing:deliver_packet/3` can be changed to possibly return that, then `hpr_gwmp_router:send/3` can return that instead of `throw({gwmp_sup_err, Reason})`.

Would that work for us?

src/hpr_routing.erl
```diff
@@ -18,7 +18,10 @@ init() ->
     ok = throttle:setup(?HOTSPOT_THROTTLE, HotspotRateLimit, per_second),
     ok.
 
--spec handle_packet(Packet :: hpr_packet_up:packet(), HandlerPid :: pid()) -> ok | {error, any()}.
+-spec handle_packet(
+    Packet :: hpr_packet_up:packet(),
+    HandlerPid :: grpcbox_stream:t()
```
ruttenberg, line 23: @michaeldjeffrey 
This type is not a pid. Below, an Error is sent to it. That can't work. For some reason, dialyzer does not complain about it.

src/protocols/hpr_gwmp_router.erl
michaeldjeffrey: That's a good point. 
Let's leave it where it is right now, and hopefully a better place will reveal itself if there is one.

src/grpc/hpr_hotspot_service.erl
```diff
@@ -21,9 +21,12 @@ init(Stream) ->
     | {stop, grpcbox_stream:t()}
     | {stop, packet_router_pb:packet_router_packet_down_v1_pb(), grpcbox_stream:t()}
     | grpcbox_stream:grpc_error_response().
-send_packet(_PacketUp, Stream) ->
+send_packet(PacketUp, Stream) ->
+    _ = hpr_routing:handle_packet(PacketUp, Stream),
```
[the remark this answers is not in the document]
michaeldjeffrey: I changed the `handle_packet/2` call to be spawned, but I think we need a bit more time to thing about how to pass the stream.

src/grpc/hpr_hotspot_service.erl
```diff
@@ -15,15 +15,13 @@ init(Stream) ->
...
-    | {ok, grpcbox_stream:t()}
-    | {ok, packet_router_pb:packet_router_packet_down_v1_pb(), grpcbox_stream:t()}
-    | {stop, grpcbox_stream:t()}
-    | {stop, packet_router_pb:packet_router_packet_down_v1_pb(), grpcbox_stream:t()}
-    | grpcbox_stream:grpc_error_response().
-send_packet(_PacketUp, Stream) ->
-    {ok, Stream}.
+    ok | {continue, grpcbox_stream:t()} | grpcbox_stream:grpc_error_response().
+send_packet(PacketUp, Stream) ->
+    _ = erlang:spawn(hpr_routing, handle_packet, [PacketUp, Stream]),
```
marcsugiyama, line 20: I think this might be hard to debug if something goes seriously wrong in hpr_routing:handle_packet/2. Maybe use `proc_lib:spawn` instead? That generates a crash report if the process exits abnormally (exit reason is not `normal`, `shutdown`, or `{shutdown, Term}`.

michaeldjeffrey (approved):

PR #27, opened 2022-08-13
successful transfer of data with CP

michaeldjeffrey:
- needed some deps
- pulled over and shredded
  - pp_udp_worker
  - pp_socket
  - semtech_udp

Only working with 1 gateway, started by hand as a local named
gen_server.

Ran `./rebar3 shell` then the following code.

```
PKB =
<<1,154,70,24,151,192,204,57,167,252,250,139,253,71,222,143,87,111,170,125,26,173,134,204,181,85,5,55,163,222,154,89,114>>.
Args = #{
         pubkeybin => PKB,
         address => "127.0.0.1",
         port => 1700
        }.
hpr_udp_worker:start_link(Args).
```

Sometimes there would be a half closed socket between HPR and gateway-rs, the cause of this is still unknown.

### DIFF
--- a/config/sys.config.src
+++ b/config/sys.config.src
@@ -1,4 +1,7 @@
 [
+    {hpr, [
+        {base_dir, "./data"}
+    ]},
     {grpcbox, [
         {servers, [
             #{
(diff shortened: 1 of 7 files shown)


michaeldjeffrey:
This served its purpose

PR #29, opened 2022-08-15
Docker support for POC

macpie:


dpezely (approved):

docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+    ports:
+      - 8081:8080
+    volumes:
+      - /var/data/hpr:/var/data
+      - /var/log/hpr:/var/log/hpr
+  promtail:
+    image: grafana/promtail
+    volumes:
+      - /var/log:/var/log
+      - /var/data/hpr/promtail-config.yaml:/etc/promtail/promtail-config.yaml
```
dpezely, line 19: Are you sure that you want a **file** instead of the conventional practice of a **subdirectory** here under `Volumes`?

dpezely (approved):
One question as sanity-check but otherwise looks fine.

michaeldjeffrey (approved):

docker-compose.yaml
macpie: Yes this is just to pass the config file, FYI docker will most likely only used for testing instances

PR #30, opened 2022-08-18
forward uplink messages to router

marcsugiyama:
This PR adds code to `hpr_protocol_router` to forward uplink messages received by `hpr` to `router`. This requires the following changes:

1. Add a grpc client config for router.
2. rebar.config and Makefile changes to generate grpc files for the router client.
3. Code to decompose the `hpr_packet_up:packet()` and assemble the `router_pb:blockchain_state_channel_message_v1_pb()`
4. grpc call to router.

TODO:
1. unit tests
2. validate the mapping between the two message types - reviewers, can you help with this?


src/hpr_routing.erl
```diff
@@ -113,10 +113,10 @@ packet_type(Packet) ->
         <<?JOIN_REQUEST:3, _:5, AppEUI:64/integer-unsigned-little,
             DevEUI:64/integer-unsigned-little, _DevNonce:2/binary, _MIC:4/binary>> ->
             {join_req, AppEUI, DevEUI};
-        (<<MType:3, _:5, DevAddr:32/integer-unsigned-little, _ADR:1, _ADRACKReq:1, _ACK:1, _RFU:1,
+        (<<FType:3, _:5, DevAddr:32/integer-unsigned-little, _ADR:1, _ADRACKReq:1, _ACK:1, _RFU:1,
```
dpezely, line 116: @mikev Which call(s) from https://github.com/helium/erlang-lorawan library should we call here instead of using Erlang's Binary syntax for descructuring-bind?

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,89 @@
 -module(hpr_protocol_router).
 
+-include("grpc/autogen/server/packet_router_pb.hrl").
+-include("grpc/autogen/client/router_pb.hrl").
+
+% TODO: should be in a common include file
+-define(JOIN_REQUEST, 2#000).
```
dpezely, line 7: Just another heads-up: Anything using this constant should instead invoke library calls from https://github.com/helium/erlang-lorawan but can be resolved in a future PR.

@mikev is the primary author of that library and should be best for recommending what to use and/or introduce new functionality.  (n.b., there's also a companion Rust library in that repo but has seen fewer miles thus far.)

dpezely (approved):
I know you mentioned having more to do such as adding tests but just wanted to add some notes along the way.  Nothing immediate.  Nothing required for this PR.

src/hpr_routing.erl
macpie: Talked with @mikev we still need to have a function in there for that

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,89 @@
...
+    wrap_grpc_response(helium_router_client:route(StateChannelMsg)).
+
+% translate hpr_packet_up:packet into blockchain_state_channel_message_v1
+-spec blockchain_state_channel_message_v1(
+    hpr_route:route(), hpr_packet_up:packet()
+) ->
+    router_pb:blockchain_state_channel_message_v1_pb().
+blockchain_state_channel_message_v1(Route, HprPacketUp) ->
+    % Decompose uplink message
+    #packet_router_packet_up_v1_pb{
```
macpie, line 27: We have a `hpr_packet_up:new/1` it's only exported for test but you can probably move that out and use it here to avoid using record directly

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,89 @@
...
+        signal_strength = SignalStrength,
+        frequency = Frequency,
+        datarate = DataRate,
+        snr = SNR,
+        region = Region,
+        hold_time = HoldTime,
+        hotspot = Hotspot,
+        signature = Signature
+    } = HprPacketUp,
+    #packet_router_route_v1_pb{
```
macpie, line 39: here you can use `hpr_route:oui/1`

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,89 @@
...
+        signal_strength = SignalStrength,
+        frequency = Frequency,
+        datarate = DataRate,
+        snr = SNR,
+        routing = RoutingInformation
+    },
+    StateChannelPacket = #blockchain_state_channel_packet_v1_pb{
+        packet = Packet,
+        hotspot = Hotspot,
+        signature = Signature,
```
macpie, line 66:
```suggestion
        signature = <<>>,
```
We might want to make the signature an empty binary as we won't be able to use it anyway

macpie:
Looking great overall, just need some unit (vs CT) probably?

marcsugiyama:
1. refactors the top level `hpr_gateway_service` code to use the `novalabsxyz` version of `grpcbox` correctly.
2. uses `grpc_client` instead of `grpcbox` as the client to talk to the old router.
3. returns the reply from the old router to the gateway.

src/protocols/hpr_protocol_router.erl
marcsugiyama: Is there a reason to avoid using the record directly? Our code is kind of a mishmash of using the functions and record directly.

src/protocols/hpr_protocol_router.erl
marcsugiyama: I wasn't sure if we'd need any of the other fields and wanted to document for myself the fields to see if they had to copied into the outgoing call parameters. Now that it's mostly done I'm guessing not? I can use the function here.

src/protocols/hpr_protocol_router.erl
marcsugiyama: Done

src/grpc/hpr_gateway_service.erl
```diff
@@ -0,0 +1,37 @@
...
+]).
+
+-spec init(atom(), StreamState :: grpcbox_stream:t()) -> grpcbox_stream:t().
+init(_RPC, StreamState) ->
+    StreamState.
+
+-spec send_packet(hpr_packet_up:packet(), grpcbox_stream:t()) ->
+    {ok, grpcbox_stream:t()} | grpcbox_stream:grpc_error_response().
+send_packet(PacketUp, StreamState) ->
+    grpcbox_service_reply(hpr_routing:handle_packet(PacketUp), StreamState).
```
michaeldjeffrey, line 20: We want to keep the spawn to `hpr_routing` here. 
If there's no destination for the packet, we just don't say anything back to the gateway.

marcsugiyama: Mack says the spawn is not needed. There is already a process per request. How does another spawn here help any?

michaeldjeffrey: You're correct. I was conflating the 2 grpc flows. :+1:

src/hpr_routing.erl
```diff
@@ -14,62 +23,63 @@
...
-handle_packet(Packet, HandlerPid) ->
-    HotspotName = hpr_utils:hotspot_name(hpr_packet_up:hotspot(Packet)),
-    lager:md([{hotspot, HotspotName}, {phash, hpr_utils:bin_to_hex(hpr_packet_up:phash(Packet))}]),
-    lager:debug("received packet from ~p", [HandlerPid]),
+-spec handle_packet(
+    Packet :: hpr_packet_up:packet()
+) -> hpr_routing_response().
+handle_packet(Packet) ->
+    GatewayName = hpr_utils:gateway_name(hpr_packet_up:gateway(Packet)),
+    lager:md([{hotspot, GatewayName}, {phash, hpr_utils:bin_to_hex(hpr_packet_up:phash(Packet))}]),
```
michaeldjeffrey, line 35:
```suggestion
    lager:md([{gateway, GatewayName}, {phash, hpr_utils:bin_to_hex(hpr_packet_up:phash(Packet))}]),
```

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
```diff
@@ -0,0 +1,292 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [IdleTimeout], []).
+
+% ------------------------------------------------------------------------------
+% gen_server callbacks
+% ------------------------------------------------------------------------------
+
+-spec init([idle_timeout()]) -> {ok, #state{}}.
+init([IdleTimeout]) ->
+    % [#connection{}]
+    ets:new(connection, [named_table, set, {keypos, #connection.lns}]),
```
michaeldjeffrey, line 78: Please use macros for ets table names.

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
```diff
@@ -0,0 +1,292 @@
...
+demonitor_owner(MonitorRef) ->
+    erlang:demonitor(MonitorRef),
+    ok.
+
+-spec decode_lns(lns()) -> endpoint().
+decode_lns(Lns) ->
+    {Address, Port} = decode_lns_parts(binary:split(Lns, <<$:>>)),
+    {tcp, Address, Port}.
+
+-spec decode_lns_parts([binary()]) -> {string(), non_neg_integer()} | none().
```
michaeldjeffrey, line 288: There's a version of this in `hpr_gwmp_router`. 
Don't know if it makes sense to reuse it here.

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
```diff
@@ -0,0 +1,292 @@
...
+            monitor_ref = MonitorRef
+        }
+    ),
+    ok.
+
+-spec set_idle_timer(lns(), idle_timeout()) -> ok.
+set_idle_timer(_Lsn, infinity) ->
+    ok;
+set_idle_timer(Lsn, IdleTimeout) ->
+    TimerRef = timer:send_after(IdleTimeout, {timeout, Lsn}),
```
michaeldjeffrey, line 266: This is an `{ok, TRef}` tuple.

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,175 @@
 -module(hpr_protocol_router).
 
+-include("grpc/autogen/server/packet_router_pb.hrl").
+
+% TODO: should be in a common include file
+-define(JOIN_REQUEST, 2#000).
```
michaeldjeffrey, line 6:
```suggestion
-include("semtech_udp.hrl").
```

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
marcsugiyama: I always make this mistake. Thanks.

src/protocols/hpr_protocol_router.erl
```diff
@@ -1,12 +1,175 @@
...
+        frequency := RX1Frequency,
+        datarate := RX1Datarate,
+        rx2_window := RX2Window
+    } = Packet,
+    % window_pb
+    #{
+        timestamp := RX2Timestamp,
+        frequency := RX2Frequency,
+        datarate := RX2Datarate
+    } = RX2Window,
```
michaeldjeffrey, line 125: RX2 won't always exist.

src/hpr_routing.erl
marcsugiyama: Done.

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
marcsugiyama: Done.

src/protocols/grpc/hpr_grpc_client_connection_pool.erl
marcsugiyama: I think we're replacing these with library functions in a lorawan utility library, once available.

src/protocols/hpr_protocol_router.erl
marcsugiyama: done

marcsugiyama: `JOIN_REQUEST` isn't in `semtech_udp.hrl`. Is the suggestion to move it there?

michaeldjeffrey: mmm, brain cross again. These are LoRa values, not Semtech UDP values. You're correct. I'm unsure where they should go right now.

src/protocols/hpr_protocol_router.erl
marcsugiyama: Done.

marcsugiyama:
This PR is updated to reflect changes to the way hpr communicates with Router. In the new design, hpr forwards gateway:send_packet uplink messages from the gateway directly to Router rather than translating the uplink to state channel messages. Similarly, the downlink messages passed from Router to the gateway.

The gateway:send_packet rpc is a bidirectional stream service. The original design defined the Router service as a unary rpc. This API difference necessitated a significant refactoring of the PR. In the new design, hpr_router_connection_manager maintains one grpc_client connection to each Router. This connection multiplexes all of the streams that need to talk to the connected Router. hpr_router_stream_manager maintains one grpc_client stream for every gateway stream that needs to talk to a Router. This one to one relationship between a gateway stream and a router stream is needed to stream answers from the Router back to the correct gateway. hpr_router_relay reads the downlink messages from the Router's grpc stream and forwards the downlink messages to the corresponding gateway's stream. grpc_client doesn't not provide an asynchronous interface. hpr_router_relay is a small gen_server that blocks on grpc_client:rcv to wait for replies from the Router.

In general, the flow is:
1. gateway opens a grpc stream to hpr
2. gateway sends an uplink message to hpr
3. hpr identifies the intended Router for the uplink message
4. hpr opens/reuses the gateway's stream to the intended Router
5. Router replies by queuing a downilnk message in the grpc stream from the Router to hpr
6. hpr reads the downlink message from the Router stream and replies to the corresponding gateway stream

marcsugiyama:
There are (a small number of) new eunit tests for the components. This code has not been tested end-to-end.

michaeldjeffrey:
We need to add some protection around Connections. 
If something dies while trying to route a packet grpcbox gets very upset and will retry until the supervisor keels over.

This probably means that we need to rethink where we account for packets.
We don't want to double account for one-off failures in routing, and we don't want to account for packets that couldn't be routed.

marcsugiyama:
I believe miner digs into the grpc_client:connection and monitors the `http_connection` process. We can do the same in the connection manager and remove the cached connection if the http process exits.

macpie:
Looks like this needs a good rebase

michaeldjeffrey:
Closed in lieu of https://github.com/helium/helium-packet-router/pull/50
Could not fix what happened to the commit history in this PR.

PR #40, opened 2022-08-31
cp multi dest

michaeldjeffrey:
The ChirpStack Application Server V4 supports multiple regions.
It does this by encoding the region into the mqtt topic the gateway-bridge publishes to.

This is a first draft solution to get _all_ regions routing to a single LNS without needing to change the Config Service/Protobuf combo.

This PR defines a completely out of udp specification `"REMAP:"` message that our udp workers know about to map one destination into another.

**Setup:**
A localhost udp_port is started that will act as a ChirpStack LNS proxy. 
It knows how to pull the Region from a `RxPk` Stat map (another out of semtech_udp spec thing we do), and look at it's setting and return a correct routing address for a gateway to send its data instead.

In normal operation, a udp_worker should only send 2 messages to this proxy before directing its data elsewhere.

The Config service should be setup to point to `locahost` and the `port` provided in the `sys.config.src` file.

**Sample Config:**
```erlang
%% /sys.config.src
{hpr, [
    {redirect_by_region, #{
        port => 1777,
        remap => #{
            <<"US915">> => <<"127.0.0.1:1778">>,
            <<"EU868">> => <<"127.0.0.1:1779">>,
            <<"AS923_1">> => <<"127.0.0.1:1780">>
        }
    }}
]}
```

### DIFF
--- a/src/protocols/gwmp/hpr_gwmp_redirect_worker.erl
+++ b/src/protocols/gwmp/hpr_gwmp_redirect_worker.erl
@@ -1,0 +1,124 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == GWMP Redirect Worker ==
+%%
+%% This worker enables remapping source destinations on the fly. Need
+%% to route a packet, but don't know the correct destination until you
+%% know the originating region? This worker fills that need.
+%%
+%% Starts a udp socket with `port'.
+%%
+%% The worker will acknowledge `pull_data' as a courtesy.
+%%
+%% Receiving a `push_data' the worker will pull the `regi' key from
+%% the `stats' map in the payload and return a special message of
+%% where that packet should be sent instead.
+%%
+%% Unmapped regions will be logged and dropped.
+%%
+%% Usage:
+%% /sys.config.src
+%% {hpr, [
+%%     {redirect_by_region, #{
+%%         port => 1777,
+%%         remap => #{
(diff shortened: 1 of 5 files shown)


dpezely (approved):

src/protocols/gwmp/hpr_gwmp_redirect_worker.erl
```diff
@@ -0,0 +1,124 @@
...
+            Region = maps:get(<<"regi">>, Stat),
+            case maps:get(Region, ReMap, undefined) of
+                undefined ->
+                    lager:warning([{region, Region}], "dropping redirect for unmapped region"),
+                    dropped;
+                NewDest ->
+                    lager:debug([{region, Region}], "redirecting gateway"),
+                    NewData = jsx:encode(#{
+                        new_dest => NewDest,
+                        packet => erlang:binary_to_list(IncomingData)
```
ruttenberg, line 102: @michaeldjeffrey 
Why is IncomingData converted to a list?

michaeldjeffrey: jsx tries to treat all binaries as strings when encoding/decoding. 
The `IncomingData` binary is raw data, and most likely doesn't sit wholly in the utf-8 space.

Converting it to a list gives us an array of the code points that jsx will ignore. 
Then we can turn it back into a binary on the other side without losing anything.

PR #41, opened 2022-09-01
Ruttenberg/hotspot to gateway

ruttenberg:
Update field names and types for packet up.


src/grpc/hpr_packet_up.erl
```diff
@@ -40,15 +40,16 @@ payload(Packet) ->
...
-    Packet#packet_router_packet_up_v1_pb.signal_strength.
+-spec rssi(Packet :: packet()) -> non_neg_integer() | undefined.
+rssi(Packet) ->
+    Packet#packet_router_packet_up_v1_pb.rssi.
 
--spec frequency(Packet :: packet()) -> float().
-frequency(Packet) ->
-    Packet#packet_router_packet_up_v1_pb.frequency.
+-spec frequency_mhz(Packet :: packet()) ->
+    float() | integer() | infinity | '-infinity' | nan | undefined.
```
dpezely, line 48: Frequency is a float, per conversations with Michael (on vacation).
We can omit `integer` here.
```suggestion
    float() | infinity | '-infinity' | nan | undefined.
```

dpezely (approved):
Thanks for accommodating those pesky name changes and modifications to the draft protobufs!

hwahby (approved):

src/grpc/hpr_packet_up.erl
ruttenberg: @dpezely 
The record definition has this:
         frequency_mhz = 0.0    :: float() | integer() | infinity | '-infinity' | nan | undefined, % = 5, optional

I thought it best to match the record definition.
If you think it is better to take out 'integer()' anyway, I am happy to do so.

cc/ @hwahby

dpezely: Ah, okay.  Keep it consistent with the record definition as you have it.  Thanks for clarifying.

PR #42, opened 2022-09-02
Stop sending PullData to Route after REMAP

michaeldjeffrey:
Workers were continuining to send pull_data to the redirect worker,
this was doing weird things with the pull_ack. After a dest is
redirected, we should no longer talk to the redirector.


test/hpr_gwmp_router_SUITE.erl
```diff
@@ -355,6 +355,11 @@ gateway_dest_redirect(_Config) ->
     USPacketUp = fake_join_up_packet(),
     EUPacketUp = USPacketUp#packet_router_packet_up_v1_pb{hotspot = PubKeyBin, region = 'EU868'},
```
ruttenberg, line 356: Field name 'hotspot' will change to 'gateway' when PR #41 is merged into 'main'.

ruttenberg (approved):

dpezely (approved):

PR #45, opened 2022-09-08
Add grpc client generation config

barndon:
This PR adds `grpc_client` as the framework for generating GRPC client code.


marcsugiyama (approved):

config/grpc_client_gen.config
```diff
@@ -0,0 +1,33 @@
...
+    {out_dir, "src/grpc/autogen/client"},
+    {beam_out_dir, "src/grpc/autogen/client"},
+    {keep_beams, false},
+    {create_services, false},
+    {type, client},
+    {override_gpb_defaults, true},
+    {gpb_opts, [
+        use_packages,
+        {maps, true},
+        {defs_as_proplists, true},
```
michaeldjeffrey, line 20: I'm not sure where we got this from, but if we're going to use straight `grpc_client` this setting needs to be false. When creating a stream `grpc_client_stream` assumes defs are maps, and will fail silently otherwise.

https://github.com/Bluehouse-Technology/grpc_client/blob/master/src/grpc_client_stream.erl#L229

barndon: Fixed.

macpie:
Included in https://github.com/helium/helium-packet-router/pull/49

PR #46, opened 2022-09-09
44 failed packet delivery to one route should not prevent delivery to remaining routes

ruttenberg:
hpr_routing:deliver_packet/3 can terminate prematurely when one of the Routes passed to it causes an error.

This PR catches the error and allows the remaining Routes to be processed.


src/hpr_routing.erl
```diff
@@ -82,16 +82,25 @@ deliver_packet(Packet, StreamHandler, [Route | Routes]) ->
...
+                router ->
+                    hpr_protocol_router:send(Packet, StreamHandler, Route);
+                gwmp ->
+                    hpr_gwmp_router:send(Packet, StreamHandler, Route);
+                _OtherProtocol ->
+                    lager:warning([{protocol, _OtherProtocol}], "unimplemented"),
+                    ok
+            end
+        catch
+            Class:Exception:Stacktrace ->
```
dpezely, line 98: Is `Stacktrace` the correct exception to catch?  This seems brittle if ever stacktraces get disabled (e.g., BEAM setting for performance).

Mind you, I'm not well versed in Erlang's exceptions... so not a show-stopper.

dpezely (approved):
Please add a description to PRs.

src/hpr_routing.erl
ruttenberg: Interesting question. Worth looking into.
If disabling stacktraces is a runtime thing, then the compiler should do something reasonable with code like this.

michaeldjeffrey:
Please remove the updates to `rebar.lock` and run the formatter.

src/hpr_routing.erl
```diff
@@ -82,16 +82,25 @@ deliver_packet(Packet, StreamHandler, [Route | Routes]) ->
...
+                lager:warning(
```
michaeldjeffrey, line 99:
```suggestion
                lager:error(
```

Hopefully this never happens, but if it does it should be big bold and red.

ruttenberg: Makes sense. Will do.

ruttenberg:
@michaeldjeffrey 
All of the changes to rebar.lock are caused by an ordinary build. Why don't you want them committed?

ruttenberg:
Formatter has been run.

src/hpr_routing.erl
marcsugiyama: The `system_flag` `backtrace_depth` limits the depth of stack captured in exceptions and returned by `process_info` on `current_stacktrace`. Is this what you're referring to? If this is set to `0`, you'd still get a `Stacktrace` value, but it will be an empty list.
Also, given the API for raising an error (where the stack trace is mandatory), I suspect there is always a `Stacktrace` value when catching an exception.
Doing anything else would make source code dependent on runtime settings, which is generally avoided.

src/hpr_routing.erl
ruttenberg: Done.

src/hpr_routing.erl
ruttenberg: Thanks for the clarification, @marcsugiyama .
I expected something like that.

hwahby (approved):

michaeldjeffrey:
> @michaeldjeffrey All of the changes to rebar.lock are caused by an ordinary build. Why don't you want them committed?

Fair enough. I forgot this branch was going into another branch that made the rebar.config changes.

michaeldjeffrey (approved):

PR #49, opened 2022-09-19
Config Service Integration

barndon:
This PR adds logic to retrieve routing configuration from Helium Config Service at run time.  It works like this:

HPR makes a gRPC call to `helium.config.config_service/route_updates` (served by Helium Config Service).  HPR sends a single, empty, `RoutesReqV1` message.  The Config Service responds with a stream of `RoutesResV1`.  Each `RoutesResV1` contains a complete list of all known routes.  For each `RoutesResV1`, HPR computes two lists:  one of them contains a mapping of devaddr ranges to route records and the other contains a mapping of EUI pairs to route records.  These mappings are stored as ETS tables and used to to perform the lookups needed to properly route incoming packets.  (Join requests rely on the EUI-to-route mappings, and uplink packets rely on the devaddr-to-route mappings.)

All of this work is done by a worker process called `hpr_routing_config_worker`, (the "Config Worker").  The Config Worker implements the `gen_server` behaviour.  It uses cast messages sent to itself to implement a simple state machine.  Once a connection is established, the Worker waits for data from the Config Service by making repeated calls to `grpc_client:rcv(Stream, Timeout)`.  If no data is received before the expiration of the Timeout, the Worker loops (possibly processing OTP system messages) and waits for data again.


config/sys.config
```diff
@@ -1,6 +1,15 @@
 [
     {hpr, [
-        {base_dir, "/var/data"}
+        {base_dir, "/var/data"},
```
michaeldjeffrey, line 3: Should we remove this file?
I don't think there's ever a time rebar3 will use it when there's a `.src` file available.

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+insert_route(Route) ->
+    DevaddrRecords = route_to_devaddr_rows(Route),
+    EUIRecords = route_to_eui_rows(Route),
+    true = ets:insert(?DEVADDRS_ETS, DevaddrRecords),
+    true = ets:insert(?EUIS_ETS, EUIRecords),
+    ok.
+
+-spec lookup_devaddr(Devaddr :: non_neg_integer()) -> list(hpr_route:route()).
+lookup_devaddr(Devaddr) ->
+    {ok, NetID} = lora_subnet:parse_netid(Devaddr, big),
```
michaeldjeffrey, line 55: Note, there are quite a few malformed devaddrs flying around. 
Parsing the NetID fails often.

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+            ],
+            ['$4']
+        }
+    ],
+    ets:select(?DEVADDRS_ETS, MS).
+
+-spec lookup_eui(AppEUI :: non_neg_integer(), DevEUI :: non_neg_integer()) ->
+    list(hpr_route:route()).
+lookup_eui(AppEUI, 0) ->
+    Routes = ets:lookup(?EUIS_ETS, {AppEUI, 0}),
```
michaeldjeffrey, line 71: I think we have this backwards. 
We want to wildcard over the `AppEUI`.

michaeldjeffrey:
Looks good

src/grpc/hpr_route.erl
```diff
@@ -3,6 +3,7 @@
 -include("../grpc/autogen/server/config_pb.hrl").
 
 -export([
+    new/1,
```
marcsugiyama, line 6: Add unit test for `new/1`?

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+    true = ets:delete(?DEVADDRS_ETS),
+    true = ets:delete(?EUIS_ETS),
+    ok.
+
+-spec update_routes(map()) -> ok.
+update_routes(#{routes := Routes}) ->
+    {NewDevaddrRows, NewEUIRows} = lists:foldl(
+        fun(RouteConfig, {DevaddrRowsAcc, EUIRowsAcc}) ->
+            Route = hpr_route:new(RouteConfig),
+            DevaddrRowsAcc2 = DevaddrRowsAcc ++ route_to_devaddr_rows(Route),
```
marcsugiyama, line 32: Do we expect the list of routes to be long? It might be more efficient to build up a list of lists and then use `lists:append` to flatten after the `foldl`.

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+-export([
+    init/0,
+    stop/0,
+    update_routes/1,
+    insert_route/1,
+    lookup_devaddr/1,
+    lookup_eui/2
+]).
+
+-define(DEVADDRS_ETS, hpr_config_routes_by_devaddr).
```
marcsugiyama, line 12: Since the keys are different for dev addr entries and eui entries, do we need two ETS tables? The shape of the key distinguishes them. Or maybe define a record for the key?

A comment describing the row format of the tables would be helpful.

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+    true = ets:insert(?DEVADDRS_ETS, DevaddrRecords),
+    true = ets:insert(?EUIS_ETS, EUIRecords),
+    ok.
+
+-spec lookup_devaddr(Devaddr :: non_neg_integer()) -> list(hpr_route:route()).
+lookup_devaddr(Devaddr) ->
+    {ok, NetID} = lora_subnet:parse_netid(Devaddr, big),
+    MS = [
+        {
+            {{'$1', '$2', '$3'}, '$4'},
```
marcsugiyama, line 58: This could be made more efficient by moving `NetID` into the pattern match, e.g.,:
```
{   
    {{NetID, '$2', '$3'}, '$4'},
    [   
        {'andalso', {'=<', '$2', Devaddr}, {'=<', Devaddr, '$3'}}
    ],  
    ['$4']
}   
```

src/hpr_config.erl
```diff
@@ -0,0 +1,178 @@
...
+    ?EUIS_ETS = ets:new(?EUIS_ETS, [public, named_table, bag, {read_concurrency, true}]),
+    ok.
+
+-spec stop() -> ok.
+stop() ->
+    true = ets:delete(?DEVADDRS_ETS),
+    true = ets:delete(?EUIS_ETS),
+    ok.
+
+-spec update_routes(map()) -> ok.
```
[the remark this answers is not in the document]
marcsugiyama: `routes_req_v1_map()` is local to `hpr_config_worker`. I think this is a `client_config_pb:routes_res_v1_pb()`?

src/hpr_config_worker.erl
```diff
@@ -0,0 +1,159 @@
...
+}).
+
+-type config_worker_opts() :: #{
+    enabled := boolean(),
+    host := string(),
+    port := integer(),
+    dets_backup_path => string()
+}.
+
+-type routes_req_v1_map() :: #{
```
marcsugiyama, line 40: Is this the same as `client_config_pb:routes_res_v1_pb()`?

michaeldjeffrey:
What do we want to do when proto updates, and loading from dets crashes.

We can:
- rename the dets table and start fresh
- blow away the dets file and not care
- try and provide some sort of upconversion

src/grpc/hpr_route.erl
macpie: done

src/hpr_config.erl
macpie: crashing you mean?

michaeldjeffrey: It will crash here with a `badmatch` if the NetID is malformed.

src/hpr_config.erl
```diff
@@ -0,0 +1,184 @@
...
+            ],
+            ['$4']
+        }
+    ],
+    ets:select(?DEVADDRS_ETS, MS).
+
+-spec lookup_eui(AppEUI :: non_neg_integer(), DevEUI :: non_neg_integer()) ->
+    list(hpr_route:route()).
+lookup_eui(AppEUI, 0) ->
+    Routes = ets:lookup(?EUIS_ETS, {AppEUI, 0}),
```
marcsugiyama, line 71: Use `ets:lookup_element(Tab, Key, 2)` instead? Then we don't need to remove the key from the results.

marcsugiyama: NM. `ets:lookup_element` throws an error if the `Key` isn't there, so this probably isn't a good API to use here.

src/hpr_config.erl
```diff
@@ -0,0 +1,184 @@
...
+
+    ExpectedDevaddrRows = lists:sort(route_to_devaddr_rows(Route)),
+    ExpectedEUIRows = lists:sort(route_to_eui_rows(Route)),
+
+    GotDevaddrRows = lists:sort(ets:tab2list(?DEVADDRS_ETS)),
+    GotEUIRows = lists:sort(ets:tab2list(?EUIS_ETS)),
+
+    ?assertEqual(ExpectedDevaddrRows, GotDevaddrRows),
+    ?assertEqual(ExpectedEUIRows, GotEUIRows),
+    ok = stop(),
```
marcsugiyama, line 136: Maybe use a `setup` fixture to call `init/0` and `stop/0` so cleanup happens even if a test asserts.

src/hpr_config_worker.erl
```diff
@@ -0,0 +1,170 @@
...
+    ok = cfg_update(),
+    {noreply, State#state{stream = Stream}}.
+
+handle_call(Msg, _From, State) ->
+    {stop, {unimplemented_call, Msg}, State}.
+
+handle_cast(Msg, State) ->
+    {stop, {unimplemented_cast, Msg}, State}.
+
+handle_info(?RCV_CFG_UPDATE, #state{connection = Connection, stream = Stream} = State) ->
```
marcsugiyama, line 107: This could also be done with a `handle_continue`. Then we don't need to msg the process.

Another possibility is to refactor this as a `gen_statem` and `next_event`, since we're basically using the first arg to `handle_continue` as the state.

src/hpr_config_worker.erl
```diff
@@ -0,0 +1,170 @@
...
+process_routes_update(RoutesResV1, State) ->
+    ok = hpr_config:update_routes(RoutesResV1),
+    maybe_cache_response(RoutesResV1, State),
+    ok.
+
+-spec maybe_cache_response(RoutesResV1 :: routes_req_v1_map(), #state{}) -> ok.
+maybe_cache_response(_RoutesResV1, #state{dets = undefined}) ->
+    ok;
+maybe_cache_response(RoutesResV1, #state{dets = Dets}) ->
+    dets:insert(Dets, {last_update, RoutesResV1}).
```
marcsugiyama, line 170: The `dets` file only ever has one row. Can't we use `file:write_file/2` and `file:read_file/1` with `term_to_binary/1` and `binary_to_term/1` instead?

marcsugiyama:
@michaeldjeffrey I'm assuming that the routes saved in `dets` is to have some routes available in case the config service is not accessible. Otherwise, it is a minor optimization, that is, when `hpr` starts it requests and receives that latest routes. Given that it's for fault tolerance, I feel we might want to do our best to use the routes in `dets`. The main challenge is probably changes to the route structure/message. We could save a version tag with the persisted routing data and then use a `code_change` like function to provide the upgrade path between the saved version and the version used by the running `hpr`.

Separately, something else to consider is to avoid starting the grpc listener until there is routing information. My thinking is that as there will be a fleet of `hpr` instances, a fresh addition to the fleet won't immediately have routing information while the others already running in the fleet have routes loaded. The new `hpr` shouldn't make itself available to load balancers until it's received some routing information.

src/grpc/hpr_route.erl
```diff
@@ -22,49 +23,49 @@
...
 -export_type([route/0]).
 
+new(#{
+    net_id := NetID,
+    devaddr_ranges := DevAddrRanges,
+    euis := EUIs,
+    protocol := Protocol,
+    oui := OUI
+}) ->
+    new(NetID, DevAddrRanges, EUIs, OUI, Protocol).
```
michaeldjeffrey, line 33: If we wanted to rely on proto for the transformations...

```erlang
config_pb:decode_msg(
  client_config_pb:encode_msg(RouteMap, route_v1_pb), 
  config_route_v1_pb
).
```

src/grpc/hpr_route.erl
```diff
@@ -127,40 +128,85 @@ new_test() ->
...
 lns_test() ->
-    Route = new(1, [{1, 10}, {11, 20}], [{1, 1}, {2, 0}], <<"lsn.lora.com>">>, gwmp, 10),
+    Route = new(
+        1,
+        [#{start_addr => 1, end_addr => 10}, #{start_addr => 11, end_addr => 20}],
+        [#{app_eui => 1, dev_eui => 1}, #{app_eui => 2, dev_eui => 0}],
+        10,
+        {gwmp, #{ip => <<"lsn.lora.com>">>, port => 80}}
+    ),
     %% If no port is in the Endpoint, 80 is assumed.
```
michaeldjeffrey, line 185: This comment no longer applies.

```suggestion
```

test/hpr_config_worker_SUITE.erl
```diff
@@ -0,0 +1,122 @@
...
+    },
+    Routes = [Route1, Route2],
+    ConfigRouteResV1 = #config_routes_res_v1_pb{routes = Routes},
+    ok = test_config_service:config_route_res_v1(ConfigRouteResV1),
+
+    %% Let  process new routes
+    timer:sleep(100),
+
+    %% Check Dets
+    {state, _Host, _Port, _Conn, _Stream, _Path, Dets} = sys:get_state(hpr_config_worker),
```
marcsugiyama, line 102: This is kind of ugly. Though I guess we'll get a match error if the state record changes. Since we know the name of the dets file, can the test open the file instead of reusing the dets handle from the worker?

test/test_config_service.erl
```diff
@@ -0,0 +1,45 @@
...
+    route_updates/2
+]).
+-export([
+    config_route_res_v1/1
+]).
+
+-spec init(atom(), StreamState :: grpcbox_stream:t()) -> grpcbox_stream:t().
+init(_RPC, StreamState) ->
+    Self = self(),
+    ok = persistent_term:put(?MODULE, self()),
```
marcsugiyama, line 19: Alternative is to register a name for `self()`.

src/hpr_config.erl
macpie: I think it should be few hundred items probably

src/hpr_config.erl
macpie: I was thinking of performance lookups wouldn't it be better with 2 tables?

src/hpr_config.erl
marcsugiyama: My concern is that `++` here is O(n^2)ish, so if we expect that there are a lot of routes, we should avoid `++`. Alternatively, we could insert to `ets` along the way, rather than accumulating a large list to insert all at once.

macpie: fair enough 👍🏻

src/hpr_config.erl
marcsugiyama: `ets` is pretty fast and we've set the `read_conucrrency` option. Having two tables makes the insert/update code a little more complicated because we have to separate out the two types of values. On the other hand, we want to optimize reads since the updates aren't expected to be that frequent(?).

marcsugiyama: `ets` is pretty fast and we've set the `read_conucrrency` option. Having two tables makes the insert/update code a little more complicated because we have to separate out the two types of values. On the other hand, we want to optimize reads since the updates aren't expected to be that frequent(?).

marcsugiyama (dismissed):

config/sys.config.src
```diff
@@ -1,6 +1,10 @@
 [
     {hpr, [
-        {base_dir, "/var/data"}
+        {config_worker, #{
+            host => "${HPR_CONFIG_WORKER_HOST}",
+            port => ${HPR_CONFIG_WORKER_PORT},
```
michaeldjeffrey, line 5: Does the relx or whatever does the config parsing support the default subst stuff yet?

```
${HPR_CONFIG_WORKER_PORT:-8080}
```

The config worker expects a string sometimes. Was this meant to be a string?

src/hpr_config.erl
```diff
@@ -0,0 +1,198 @@
...
+-define(DEVADDRS_ETS, hpr_config_routes_by_devaddr).
+-define(EUIS_ETS, hpr_config_routes_by_eui).
+
+-spec init() -> ok.
+init() ->
+    ?DEVADDRS_ETS = ets:new(?DEVADDRS_ETS, [public, named_table, bag, {read_concurrency, true}]),
+    ?EUIS_ETS = ets:new(?EUIS_ETS, [public, named_table, bag, {read_concurrency, true}]),
+    ok.
+
+-spec stop() -> ok.
```
michaeldjeffrey, line 21: `stop/0` looks like test only.

config/sys.config.src
macpie: it does will add default port tho

test/hpr_config_worker_SUITE.erl
```diff
@@ -0,0 +1,126 @@
...
+
+    %% Let  process new routes
+    timer:sleep(100),
+
+    %% Check backup file
+    FilePath = proplists:get_value(file_backup_path, Config),
+    case file:read_file(FilePath) of
+        {ok, Binary} ->
+            #{routes := DetsRoutes} = erlang:binary_to_term(Binary),
+            ?assertEqual(Routes, [hpr_route:new(R) || R <- DetsRoutes]);
```
michaeldjeffrey, line 109:
```suggestion
            #{routes := BackupRoutes} = erlang:binary_to_term(Binary),
            ?assertEqual(Routes, [hpr_route:new(R) || R <-BackupRoutes]);
```

src/hpr_config_worker.erl
```diff
@@ -0,0 +1,165 @@
...
+            {noreply, State, {continue, ?RCV_CFG_UPDATE}};
+        eof ->
+            lager:warning("got eof"),
+            _ = grpc_client:stop_connection(Connection),
+            {noreply, State, {continue, ?CONNECT}};
+        {error, timeout} ->
+            {noreply, State, {continue, ?RCV_CFG_UPDATE}};
+        {error, E} ->
+            lager:error("failed to rcv ~p", [E]),
+            {stop, {error, E}}
```
michaeldjeffrey, line 115: The main supervisor is a `rest_for_one`, and this worker is second.
If we go down when this connection closes, we're going to close _all_ connections.

Same with the `stop`s in `handle_call` and `handle_cast`.

marcsugiyama: Should the main supervisor be `one_for_one`?

michaeldjeffrey (approved):

PR #55, opened 2022-09-27
GRPC update

michaeldjeffrey:
Gets us back to a working state with gateway-rs and the new `packet_router.proto` definitions.

`config.proto` was broken out of the packet_router namespace into its own.
For the interim, it is being compiled in the server file because we quite heavily use the records generated in `hpr_route.erl`.

There's a PR coming for getting talking to the Config service as a client, we can figure out the whole maps vs records debacle in that PR.


test/hpr_protocol_router_SUITE.erl
```diff
@@ -304,19 +302,31 @@ grpc_full_flow_connection_refused_test(_Config) ->
 %% ===================================================================
 
 test_packet() ->
-    {packet_router_packet_up_v1_pb,
+    {
```
marcsugiyama, line 305: Should this use the record syntax instead of constructing the tuple directly?

marcsugiyama (approved):
Looks good. As noted, if we use `grpc_client`, the config message uses maps instead of records. It doesn't look like `hpr_routing_config_worker` will need to change through.

test/hpr_protocol_router_SUITE.erl
macpie: I am fixing in another PR

macpie (approved):

PR #60, opened 2022-09-30
make sure metadata for logging is all strings

michaeldjeffrey:
Promtail doesn't like some types of fields. And it happens to be binaries.


macpie (approved):

marcsugiyama (approved):
I'm not that familiar with `lager`. What problem is this solving? Is the `gateway_mac` field sometimes not formatted properly?

michaeldjeffrey:
> I'm not that familiar with `lager`. What problem is this solving? Is the `gateway_mac` field sometimes not formatted properly?

There is a promtail container following the logs. It expects the `field=msg` format for it's parsing.
It doesn't like binary fields, so it complains about things that look like `gateway_mac=<<"asdf">>`.

PR #61, opened 2022-10-03
Basic metrics

macpie:


src/grpc/hpr_packet_up.erl
```diff
@@ -119,6 +128,37 @@ to_map(PacketRecord) ->
...
+            DevEUI:64/integer-unsigned-little, _DevNonce:2/binary, _MIC:4/binary>> ->
+            {join_req, {AppEUI, DevEUI}};
+        (<<FType:3, _:5, DevAddr:32/integer-unsigned-little, _ADR:1, _ADRACKReq:1, _ACK:1, _RFU:1,
+            FOptsLen:4, _FCnt:16/little-unsigned-integer, _FOpts:FOptsLen/binary,
+            PayloadAndMIC/binary>>) when
+            (FType == ?UNCONFIRMED_UP orelse FType == ?CONFIRMED_UP) andalso
+                %% MIC is 4 bytes, so the binary must be at least that long
+                erlang:byte_size(PayloadAndMIC) >= 4
+        ->
+            Body = binary:part(PayloadAndMIC, {0, byte_size(PayloadAndMIC) - 4}),
```
marcsugiyama, line 144: Since we're not looking at any value past the `Port`, I wonder if this would be clearer if we check the length and then pattern match, something like:
```
if
    byte_size(PayloadAndMIC) == 4 -> undefined;
    true -> <<Port:8, _/binary>>, Port
end
```

src/hpr_routing.erl
```diff
@@ -28,31 +22,37 @@ init() ->
...
 ) -> hpr_routing_response().
 handle_packet(Packet) ->
+    Start = erlang:system_time(millisecond),
     GatewayName = hpr_utils:gateway_name(hpr_packet_up:gateway(Packet)),
     lager:md([{gateway, GatewayName}, {phash, hpr_utils:bin_to_hex(hpr_packet_up:phash(Packet))}]),
-    %% TODO: log some identifying information?
-    lager:debug("received packet"),
+    PacketType = hpr_packet_up:type(Packet),
+    {Type, _} = PacketType,
+    lager:debug("received ~p packet", [Type]),
```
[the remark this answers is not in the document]
macpie: good idea

src/metrics/hpr_metrics.erl
```diff
@@ -52,6 +76,15 @@ handle_cast(_Msg, State) ->
 
 handle_info(?METRICS_TICK, State) ->
     lager:debug("running metrics"),
+    _ = erlang:spawn_opt(
```
marcsugiyama, line 79: A minor concern here, is if `record_grpc_connections` get stuck or is slow, we could spawn off a lot of these. Consider some code to make sure there is only one of these running at a time? We had something similar to this at my last company. When the system got bogged down, the metrics collection occasionally took longer than interval between collections which lead to additional load, just when we didn't need it, and confusing metrics.

src/metrics/hpr_metrics.erl
```diff
@@ -52,6 +76,15 @@ handle_cast(_Msg, State) ->
...
+        fun() ->
+            ok = record_grpc_connections()
+        end,
+        [
+            {fullsweep_after, 0},
```
marcsugiyama, line 84: Do we need these options? Setting priority to high seems like premature optimization, unless you have prior experience that this is needed. Given that this process is short lived(?), do we really need to do a full sweep after minor GCs? This also feels like premature optimization.

src/metrics/hpr_metrics.erl
marcsugiyama: NM - it's not an interval send by send after.

macpie: Good note actually saw this once.

src/metrics/hpr_metrics.erl
macpie: yeah might be much right now, we used to have it before in Router simple copy/paste here tbh

src/metrics/hpr_metrics.erl
```diff
@@ -34,12 +35,35 @@
...
 %% ------------------------------------------------------------------
 init(_Args) ->
     erlang:process_flag(trap_exit, true),
     lager:info("init"),
+    ElliOpts = [
+        {callback, hpr_metrics_handler},
+        {callback_args, #{}},
+        {port, 3000}
+    ],
+    {ok, _Pid} = elli:start_link(ElliOpts),
```
marcsugiyama, line 66: Should this be started by the supervisor? It's somewhat independent of the this gen_server, isn't it?

michaeldjeffrey: We do this in Packet Purchaser in the top level sup for reference.
https://github.com/helium/packet-purchaser/blob/master/src/pp_sup.erl#L93-L100

macpie: 👍🏻

src/metrics/hpr_metrics.erl
marcsugiyama: Right - the send after is outside the spawn, so there could be more than one running at a time.

michaeldjeffrey (approved):

PR #65, opened 2022-10-05
odds n ends

michaeldjeffrey:
- remove `hpr` from file backup path. In docker-compose we're already mounting to a nested `hpr` dir, and this creates a nested dir.
- add a flag for config to not blow away all routes if the config service sends an empty list of routes.
  - Hopefully removed in the future, but useful right now for testing with a routing file but an empty config service.
- fixup leftover eunit route creations. Not technically failures, but good to have everything consistent.

### DIFF
--- a/config/sys.config.src
+++ b/config/sys.config.src
@@ -3,7 +3,7 @@
         {config_worker, #{
             host => "${HPR_CONFIG_WORKER_HOST}",
             port => "${HPR_CONFIG_WORKER_PORT:-8080}",
-            file_backup_path => "/var/data/hpr/config_worker.backup"
+            file_backup_path => "/var/data/config_worker.backup"
         }}
     ]},
     {grpcbox, [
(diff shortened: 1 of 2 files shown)


config/sys.config.src
```diff
@@ -3,7 +3,7 @@
         {config_worker, #{
             host => "${HPR_CONFIG_WORKER_HOST}",
             port => "${HPR_CONFIG_WORKER_PORT:-8080}",
-            file_backup_path => "/var/data/hpr/config_worker.backup"
+            file_backup_path => "/var/data/config_worker.backup"
```
macpie, line 6: Why?

michaeldjeffrey: - remove hpr from file backup path. In docker-compose we're already mounting to a nested hpr dir, and this creates a nested dir.

macpie (approved):

PR #67, opened 2022-10-06
Store data_src as ip_address() for matching pull_ack Address

michaeldjeffrey:
When gen_udp handles a message it provides the address as a tuple of numbers. We stored it as the string we received, so we were missing token acks.

### DIFF
--- a/src/protocols/gwmp/hpr_gwmp_worker.erl
+++ b/src/protocols/gwmp/hpr_gwmp_worker.erl
@@ -52,7 +52,8 @@
     pull_data = #{} :: pull_data_map(),
     pull_data_timer :: non_neg_integer(),
     shutdown_timer :: {Timeout :: non_neg_integer(), Timer :: reference()},
-    dest_remap = #{} :: #{socket_dest() => socket_dest()}
+    dest_remap = #{} :: #{socket_dest() => socket_dest()},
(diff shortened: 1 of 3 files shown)


macpie (approved):

src/protocols/gwmp/hpr_gwmp_worker.erl
```diff
@@ -237,7 +239,7 @@ handle_udp(
                         State0;
                     {DestMap1, Resend} ->
                         {{A, B, C, D}, Port} = DataSrc,
-                        Key = {lists:flatten(io_lib:format("~p.~p.~p.~p", [A, B, C, D])), Port},
+                        Key = {{A, B, C, D}, Port},
```
marcsugiyama, line 242: Isn't this `Key = DataSrc`? Do we need `Key`?

marcsugiyama: The address pattern match only matches ipv4 addresses. Do we care about ipv6 addresses?

michaeldjeffrey: That it is. It is not needed.

I think that restriction only applies to this branch where we're remapping a destination based on region. 
I'm hoping we can drop this code before long in lieu of a better solution.

If it's easy enough to support ipv6 here, and it won't fail silently, by all means let's do that.

PR #74, opened 2022-10-10
Max Copies

macpie:


michaeldjeffrey:
Wondering if we want to get rid of the `_buy` terminology?

Also, we need to allow for different routes to want a different number of packets.

macpie:
> Wondering if we want to get rid of the `_buy` terminology?
> 
> Also, we need to allow for different OUIs to want a different number of packets.

It's based per route

src/hpr_routing.erl
```diff
@@ -92,25 +86,31 @@ deliver_packet(Packet, [Route | Routes]) ->
...
-            {gwmp, _} ->
-                hpr_protocol_gwmp:send(Packet, self(), Route);
-            _OtherProtocol ->
-                lager:warning([{protocol, _OtherProtocol}], "unimplemented"),
-                ok
-        end,
-    case Resp of
+    Key = crypto:hash(sha256, <<
+        (hpr_packet_up:phash(Packet))/binary, (hpr_route:lns(Route))/binary
+    >>),
```
michaeldjeffrey, line 91: This answer my previous comment about allowing different numbers of packet per OUI.

I don't think it would be a widespread problem, but putting here for posterity sake because I'm thinking about it.

If for some reason, someone made multiple routes that went to the same endpoint.

e.g.
- Wildcard over the same JoinEUI, but split DevAddrs for easier activation of routing. For those joins, they would get half the amount they expected, but twice.

- If someone had a set of routes that happened to overlap DevAddr addresses (whether on purpose, or not) those would also get half the packets they were expecting, but twice.

macpie: I am thinking that we should probably add some kind of unique ID per route. It would save this problem as well as allow us to do individual updates.

macpie:
Will have to wait on https://github.com/helium/proto/pull/204 but this is dry

macpie:
Rdy now

michaeldjeffrey (approved):

src/hpr_max_copies.erl
```diff
@@ -0,0 +1,140 @@
...
+%% ------------------------------------------------------------------
+%% Internal Function Definitions
+%% ------------------------------------------------------------------
+
+-spec scheduled_cleanup(Duration :: non_neg_integer()) -> ok.
+scheduled_cleanup(Duration) ->
+    erlang:spawn(
+        fun() ->
+            Time = erlang:system_time(millisecond) - Duration,
+            Expired = select_expired(Time),
```
marcsugiyama, line 56: Can this be `ets:select_delete`?

src/hpr_max_copies.erl
```diff
@@ -0,0 +1,140 @@
...
+
+-spec init() -> ok.
+init() ->
+    ets:new(?ETS, [
+        public,
+        named_table,
+        set,
+        {write_concurrency, true}
+    ]),
+    ok = scheduled_cleanup(?CLEANUP),
```
marcsugiyama, line 23: If this crashes for some reason, then the cleanups stop. Should this be a supervised process?

src/hpr_max_copies.erl
```diff
@@ -0,0 +1,140 @@
...
+
+-define(ETS, hpr_max_copies_ets).
+-define(MAX_COPIES, max_copies).
+-define(MAX_TOO_LOW, max_copies_max_too_low).
+-define(CLEANUP, timer:hours(1)).
+-define(CLEANUP_SLEEP, timer:hours(1)).
+
+-spec init() -> ok.
+init() ->
+    ets:new(?ETS, [
```
marcsugiyama, line 17: Document the format of a row?
How do values get in the table?

marcsugiyama: Oh, I see how values get in. I hadn't realized `update_counter` adds missing rows.

test/hpr_routing_SUITE.erl
```diff
@@ -118,5 +160,110 @@ join_req_test(_Config) ->
...
+
+    #{secret := PrivKey2, public := PubKey2} = libp2p_crypto:generate_keys(ecc_compact),
+    SigFun2 = libp2p_crypto:mk_sig_fun(PrivKey2),
+    Gateway2 = libp2p_crypto:pubkey_to_bin(PubKey2),
+
+    UplinkPacketUp2 = test_utils:uplink_packet_up(#{
+        gateway => Gateway2, sig_fun => SigFun2, devaddr => DevAddr, fcnt => 1
+    }),
+
+    #{secret := PrivKey3, public := PubKey3} = libp2p_crypto:generate_keys(ecc_compact),
```
marcsugiyama, line 202: Maybe create a utility function to make a up link packet, to avoid this sequence of code. Might be a bit clearer what the test is doing.

marcsugiyama (dismissed):

src/hpr_max_copies.erl
michaeldjeffrey: This isn't a process. It's called by the supervisor.

marcsugiyama: Still, if the process running the cleanup were to crash for some reason, the cleanups stop running.

michaeldjeffrey: Oh, indeed. 
I misattributed the original message to the making of the ets table.

src/hpr_max_copies.erl
```diff
@@ -0,0 +1,141 @@
...
+    end),
+    ok.
+
+%% ------------------------------------------------------------------
+%% Internal Function Definitions
+%% ------------------------------------------------------------------
+
+-spec scheduled_cleanup(Duration :: non_neg_integer()) -> ok.
+scheduled_cleanup(Duration) ->
+    {ok, _} = timer:apply_interval(Duration, ?MODULE, cleanup, [Duration]),
```
marcsugiyama, line 60: Minor concern is to avoid having more than one of these running at a time should it get stuck for some reason. But the interval times are quite long, so I think we're fine here.

src/hpr_max_copies.erl
```diff
@@ -0,0 +1,141 @@
...
+        named_table,
+        set,
+        {write_concurrency, true}
+    ]),
+    ok = scheduled_cleanup(?CLEANUP_TIME),
+    ok.
+
+-spec check(Key :: binary(), Max :: non_neg_integer()) ->
+    ok | {error, ?MAX_TOO_LOW | ?MAX_COPIES}.
+check(_Key, Max) when Max =< 0 ->
```
marcsugiyama, line 29: I'm wondering if the function name could be better? To me `check` doesn't imply a state change (counting the checks). Maybe it should be `count`? or `count_with_check`?

marcsugiyama (approved):

michaeldjeffrey (approved):

PR #75, opened 2022-10-11
Upgrade RSSI to be an int

macpie:
Waiting on https://github.com/helium/proto/pull/199


michaeldjeffrey (approved):

rebar.config
```diff
@@ -13,9 +13,7 @@
     {grpcbox,
         {git, "https://github.com/novalabsxyz/grpcbox.git",
             {ref, "6180be0f4f4f199f824e60ae5c64ddd2daa5acff"}}},
-    {helium_proto,
-        {git, "https://github.com/helium/proto.git",
-            {ref, "1a1a4ae82a898620d96cb486ed76d3857d3ed8f2"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
```
ruttenberg, line 16: @macpie 
I think that specifying a branch is a very bad idea.
This should specify a commit hash or, better yet, a tag to communicate clearly what is intended.
Currently, on _main_, this is in _rebar.lock_,
       {ref,"d4dc09d89ec4e99ab6ec9992f207e33a6d9a3e30"}},
but this is the current commit hash for master:
97d063d2eeec0e950e90f50f050e8048a5031a48

PR #77, opened 2022-10-13
unknown pull ack 2

michaeldjeffrey:
Keep the fact that a pull_data was acknowledged.

When a packet was sent, the worker would check to see if a pull_data had
been sent to the dest. That map is was checking only stored in-flight
pull_data. As soon as a pull_ack came through (normally immediate), the
worker would lose the ability to know if there was already a pull_data
session in tact with a specific destination.

The result was a new 10s pull_data timer being started every time a
packet was pushed.

Now, we keep the dest in the pull_data map and mark it as acknowledged.
The next pull_data will come and overwrite the entry.


marcsugiyama (approved):
Code looks good to me. I don't know the gwmp protocol requirements, so I can't review if it's doing the right thing.

macpie (approved):

PR #84, opened 2022-10-18
Start HPR CLI

michaeldjeffrey:
- add getter for base protocol_type
- store routes received as is for CLI querying
- CLI config
  - ls
  - oui info


macpie:
Is there a way when we receive Routes to print some of it (like cli) to then be able to extract them with Loki in a table?

macpie (approved):

PR #89, opened 2022-10-27
Support new config service APIs

macpie:
Fix for https://github.com/helium/helium-packet-router/issues/86


src/hpr_sup.erl
```diff
@@ -46,6 +46,22 @@ start_link() ->
     supervisor:start_link({local, ?SERVER}, ?MODULE, []).
 
 init([]) ->
+    KeyFileName = application:get_env(hpr, key, "/var/data/hpr.key"),
+    ok = filelib:ensure_dir(KeyFileName),
+    Key =
```
michaeldjeffrey, line 51: Can we put the saving and getting of the pub_key closer together?

macpie: What do you mean?

michaeldjeffrey (approved):

PR #95, opened 2022-11-03
Handle hpr_router_connection_manager crash takes down all app

macpie:


michaeldjeffrey (dismissed):

src/protocols/router/hpr_router_connection_manager.erl
```diff
@@ -47,85 +46,66 @@ start_link() ->
...
+    lager:info("connection ~p to ~s went down ~p", [_Pid, Lns, _ExitReason]),
+    true = ets:delete(?CONNECTION_TAB, Lns),
+    {noreply, State};
+handle_info({'EXIT', _Pid, _Reason}, State) ->
+    lager:info("connection ~p exited ~p", [_Pid, _Reason]),
     {noreply, State}.
 
+terminate(_Reason, #state{}) ->
+    lager:error("terminate ~p", [_Reason]),
+    _ = ets:delete(?CONNECTION_TAB),
```
marcsugiyama, line 102: This process owns the ets table, so it will be dropped automatically when the process exits. I don't think we need this code.

src/protocols/router/hpr_router_connection_manager.erl
```diff
@@ -47,85 +46,66 @@ start_link() ->
...
 
 -spec handle_info({{'DOWN', lns()}, reference(), process, pid(), any()}, #state{}) ->
     {noreply, #state{}}.
 handle_info({{'DOWN', Lns}, _Mon, process, _Pid, _ExitReason}, State) ->
-    ets:delete(State#state.connection_table, Lns),
+    lager:info("connection ~p to ~s went down ~p", [_Pid, Lns, _ExitReason]),
+    true = ets:delete(?CONNECTION_TAB, Lns),
+    {noreply, State};
+handle_info({'EXIT', _Pid, _Reason}, State) ->
+    lager:info("connection ~p exited ~p", [_Pid, _Reason]),
```
marcsugiyama, line 97: I think it would be clearer if we add `process_flag(trap_exit, true)` to `init`, although I do see that `http2_client:new_connection` does this. I think that's a confusing side effect of creating a new `http2` connection, which makes it a little mysterious how we can get `'EXIT'` messages without setting the process flag

src/protocols/router/hpr_router_connection_manager.erl
macpie: I remove the ETS for now. I don't think there will be enough Router connection for this to matter. Note also that Router will eventually die and this code with it.

src/protocols/router/hpr_router_connection_manager.erl
marcsugiyama: Yes? But that we're getting one at all is kind of confusing if you don't know that `http2_client:new_connection` is setting the trap_exit process flag. I still think we should set it in `init`.

marcsugiyama (dismissed):

marcsugiyama (approved):

PR #97, opened 2022-11-08
Add backoff support for connecting to HCS

macpie:


src/hpr_config_worker.erl
```diff
@@ -114,11 +131,15 @@ handle_continue(
             ok = process_route_stream_res(hpr_route_stream_res:from_map(RouteStreamRes), Path),
             {noreply, State, {continue, ?RCV_CFG_UPDATE}};
         eof ->
-            lager:warning("got eof"),
-            _ = grpc_client:stop_connection(Connection),
-            {noreply, State, {continue, ?CONNECT}};
+            {Delay, Backoff1} = backoff:fail(Backoff0),
```
marcsugiyama, line 134: After this `backoff:fail/1`, we go back to `handle_continue(?CONNECT, ...)`, which means the connect is now starting with a longer backoff. I'm wondering if the calls to `backoff` are needed here. Since we don't stay in `RCV_CFG_UPDATE` after the failure, do we need to call `backoff:succeed/1`?

test/hpr_config_worker_SUITE.erl
```diff
@@ -303,6 +309,32 @@ delete_route_test(Config) ->
...
+    ServerPid = proplists:get_value(server_pid, Config),
+
+    meck:new(backoff, [passthrough]),
+
+    gen_server:stop(ServerPid),
+
+    ?assertEqual(1, meck:num_calls(backoff, fail, 1)),
+
+    %% We sleep 3s the BACKOFF_MIN is 1s so we should get another 2 tries
+    timer:sleep(3000),
```
marcsugiyama, line 326: If can assume that `backoff` works properly, you could make this test run faster by mocking `backoff` and have it return canned values. If you do this, you won't need to the `TEST` definitions for the timeouts. It might be easier to also mock `grpc_client:connect` to simulate the failure instead of stopping the server.

src/hpr_config_worker.erl
```diff
@@ -23,12 +23,20 @@
     terminate/2
 ]).
 
+-ifdef(TEST).
```
marcsugiyama, line 26: These can also be parameters to `init`.

test/hpr_config_worker_SUITE.erl
macpie: I am not a massive fan of mocking all thew time I would rather wait an extra few seconds and get the "real" behavior.

michaeldjeffrey (approved):

PR #98, opened 2022-11-09
Use hpr_packet_service:packet_down/2; fix tests.

ruttenberg:


src/grpc/packet_router/hpr_packet_down.erl
```diff
@@ -91,7 +91,7 @@ window(WindowMap) ->
 window(TS, FrequencyHz, DataRate) ->
     WindowMap = #{
         timestamp => TS,
-        frequency => FrequencyHz,
+        frequency => round(FrequencyHz),
```
dpezely, line 94: Since `round()` drops precision, please add a comment as to *why* that loss in precision is necessary-- here and the other location (below) where it gets called

dpezely (dismissed):
Approved, but adding a comment would help with future work.

macpie (approved):

config/sys.config.src
```diff
@@ -14,9 +14,10 @@
             port => "${HPR_CONFIG_WORKER_PORT}",
             file_backup_path => "/var/data/config_worker.backup"
         }},
-        {http_roaming_downlink_port, 8090}
+        {http_roaming_downlink_port, 8085}
     ]},
     {grpcbox, [
+        {listen_opts, #{port => 8083, ip => {0, 0, 0, 0}}},
```
michaeldjeffrey, line 20: testing ports?

src/protocols/http/hpr_http_roaming_downlink_handler.erl
```diff
@@ -92,5 +92,5 @@ handle_event(Event, _Data, _Args) ->
 ) -> ok.
 send_response(ResponseStream, DownlinkPacket) ->
     lager:debug("sending response: ~p, pid: ~p", [DownlinkPacket, ResponseStream]),
```
michaeldjeffrey, line 94: This wasn't changed, but I don't think we should be logging the whole downlink packet here. 
Log aggregator getting upset about binaries and whatnot.

It might be nice to have a log message about a downlink being sent from which protocol.

src/grpc/packet_router/hpr_packet_down.erl
ruttenberg: I will add a comment.

config/sys.config.src
ruttenberg: I will revert the ports to the original values.

ruttenberg: Done.

src/protocols/http/hpr_http_roaming_downlink_handler.erl
ruttenberg: Done.

src/grpc/packet_router/hpr_packet_down.erl
```diff
@@ -91,7 +91,10 @@ window(WindowMap) ->
 window(TS, FrequencyHz, DataRate) ->
     WindowMap = #{
         timestamp => TS,
-        frequency => FrequencyHz,
+
+        %%        the protobuf encoding requires that the frequency is an integer, rather than a float in exponential notation
+        frequency => round(FrequencyHz),
```
dpezely, line 96: Considering context of the new code comment:

Perhaps there should be a **multiplier**-- rather than `round`, `trunc` or `floor` here.

The rationale for using integers for Frequency is to avoid errors due to different encodings, because "float" means different things on different architectures.

There's a new `Contributing` section in the README for https://github.com/helium/proto repo highlighting this issue.

ruttenberg: @dpezely 
Frequency in the protobuf is uint32, so I think that the  guideline is being followed.
In the roaming protocol, frequency is expressed as Mhz, in a float.
I don't see how the conversion can be avoided.
cc/ @michaeldjeffrey

michaeldjeffrey: This probably was brought forth by the HTTP Roaming Interface where frequencies are floats in JSON payloads.

When making constructing our protobuf that expects an integer, we multiply to get to Hertz.

```erlang
> 923.5 * 1_000_000.
9.235e8
```

Though it could be an integer, it's still be represented by a float.

michaeldjeffrey (approved):

macpie (approved):

src/grpc/packet_router/hpr_packet_down.erl
dpezely: Going from hz to Mhz, wouldn't the calculation be `round(FrequencyHz * 1000)` instead of `round(FrequencyHz)`?

Otherwise, a plan such as EU434a with frequency range of 433.175 .. 434.665 would lose all of its channels due to that loss of precision.
https://github.com/helium/erlang-lorawan/blob/master/src/lora_plan.erl#L792-L793

michaeldjeffrey: We're not turning into Mhz. 
The function and the proto expect Hz.
This is ensuring that with `round/`.

`round(FrequencyHz * 1000)` would get us Khz.

Issue #108, opened 2022-11-29
Add hotspot metadata in uplinks


lytzV:
We will need to add hotspot name in our uplinks & join accepts so that Chirpstack can display it in LoRaWAN frames & Events.

https://github.com/chirpstack/chirpstack-gateway-bridge/blob/master/internal/backend/semtechudp/packets/push_data.go#L345

lytzV:
Will this mess up other UDP based ingestion if they don't parse for "meta"?

macpie:
I don't think it should

PR #111, opened 2022-12-01
Add meta field to gwmp rxpk

macpie:
Fix for #108

### DIFF
--- a/src/grpc/packet_router/hpr_packet_router_service.erl
+++ b/src/grpc/packet_router/hpr_packet_router_service.erl
@@ -29,6 +29,7 @@ route(EnvUp, StreamState) ->
             {ok, StreamState};
         {register, Reg} ->
             PubKeyBin = hpr_register:gateway(Reg),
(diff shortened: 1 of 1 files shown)


lytzV (dismissed):

src/protocols/hpr_protocol_gwmp.erl
```diff
@@ -81,15 +79,20 @@ packet_up_to_push_data(Up, GatewayTime) ->
...
             chan => 0,
-
             datr => erlang:atom_to_binary(hpr_packet_up:datarate(Up)),
             rssi => hpr_packet_up:rssi(Up),
             lsnr => hpr_packet_up:snr(Up),
             size => erlang:byte_size(hpr_packet_up:payload(Up)),
-            data => base64:encode(hpr_packet_up:payload(Up))
+            data => base64:encode(hpr_packet_up:payload(Up)),
+            meta => #{
+                gateway_id => B58,
```
michaeldjeffrey, line 88: I think we need to cast this to a binary to make sure it get's encoded as a string and not a list.

michaeldjeffrey (approved):

PR #117, opened 2022-12-07
Include new downlink service

macpie:


michaeldjeffrey (dismissed):
Probably we should do something with the response codes from `process_downlink`. But I'm not sure what yet, that will probably become clearer later.

michaeldjeffrey (approved):

michaeldjeffrey (approved):

PR #123, opened 2022-12-12
Add no_routes options

macpie:


michaeldjeffrey (approved):

src/hpr_routing.erl
```diff
@@ -36,11 +38,18 @@ handle_packet(Packet) ->
...
         ok ->
-            Routes = find_routes(PacketType),
-            lager:debug("maybe deliver packet to ~w routes", [erlang:length(Routes)]),
-            ok = maybe_deliver_packet(Packet, Routes),
-            hpr_metrics:observe_packet_up(PacketType, ok, erlang:length(Routes), Start),
-            ok
+            case find_routes(PacketType) of
+                [] ->
+                    lager:debug("no routes found"),
+                    ok = maybe_delvier_no_routes(Packet),
```
michaeldjeffrey, line 44: We know the protocol. Do we want to directly send them here?
Otherwise, every packet will be sent to reporting.

michaeldjeffrey (approved):

PR #124, opened 2022-12-15
Rename of config service to iot_config

macpie:


rebar.config
```diff
@@ -67,10 +67,10 @@
 ]}.
 
 {xref_ignores, [
-    config_pb,
+    iot_config_pb,
     packet_router_pb,
     client_packet_router_pb,
-    client_config_pb
+    client_iot_config_pb
```
michaeldjeffrey, line 73: It won't let me put a suggestion. 
We don't generate the `client_` services anymore. Those can be removed.

michaeldjeffrey (approved):

PR #127, opened 2022-12-20
Add more stats

macpie:


src/cli/hpr_cli_trace.erl
```diff
@@ -0,0 +1,85 @@
...
+        [trace_cmd()]
+    ).
+
+%%--------------------------------------------------------------------
+%% Config
+%%--------------------------------------------------------------------
+
+trace_usage() ->
+    [
+        ["trace"],
```
michaeldjeffrey, line 35: I just remembered this trace from blockchain-core.
https://github.com/helium/blockchain-core/blob/master/src/cli/blockchain_cli_trace.erl

```sh
hpr trace start app_eui=aabbcc dev_eui=ddeeff
```
And it dumps to stdout if a file isn't specified. Might be useful in the future.

michaeldjeffrey (approved):

src/cli/hpr_cli_trace.erl
macpie: This one just creates the file by default in trance function

PR #133, opened 2023-01-18
Macpie/proto scale routes

macpie:


src/cli/hpr_cli_config.erl
```diff
@@ -105,15 +105,16 @@ config_oui_list(["config", "oui", OUIString], [], []) ->
...
         Server = hpr_route:server(Route),
 
-        DevAddrRanges = lists:map(FormatDevAddr, hpr_route:devaddr_ranges(Route)),
+        DevAddrRanges = lists:map(
+            FormatDevAddr, hpr_route_ets:devaddr_ranges_for_route(hpr_route:id(Route))
+        ),
         DevAddrInfo = [DevAddrHeader | DevAddrRanges],
 
-        EUIs = lists:map(FormatEUI, hpr_route:euis(Route)),
+        EUIs = lists:map(FormatEUI, hpr_route_ets:eui_pairs_for_route(hpr_route:id(Route))),
```
michaeldjeffrey, line 113: If this is going to be useful, we might to put this behind a flag. 
This will easily be a few thousand for router operator Routes.

src/grpc/iot_config/hpr_route_stream_worker.erl
```diff
@@ -150,8 +150,12 @@ handle_info(?INIT_STREAM, #state{conn_backoff = Backoff0} = State) ->
     end;
 %% GRPC stream callbacks
 handle_info({data, _StreamID, RouteStreamRes}, #state{file_backup_path = Path} = State) ->
-    lager:debug("route update"),
-    ok = process_route_stream_res(RouteStreamRes, Path),
+    Action = hpr_route_stream_res:action(RouteStreamRes),
+    Data = hpr_route_stream_res:data(RouteStreamRes),
+    {Type, _} = Data,
+    lager:debug("got route stream update ~s ~s ", [Action, Type]),
```
michaeldjeffrey, line 156:
```suggestion
    lager:debug([{action, Action}, {type, Type}], "got route stream update"),
```

So we can query them with Loki if we want to.

src/grpc/iot_config/hpr_route_stream_worker.erl
```diff
@@ -205,40 +209,96 @@ terminate(_Reason, _State) ->
...
+    RouteStreamRes :: hpr_route_stream_res:action(),
+    Data ::
+        {route, hpr_route:route()}
+        | {eui_pair, hpr_eui_pair:eui_pair()}
+        | {devaddr_range, hpr_devaddr_range:devaddr_range()}
+) -> ok.
+maybe_cache_response(undefined, _Action, _Data) ->
     ok;
-maybe_cache_response(RouteStreamRes, Path) ->
+maybe_cache_response(Path, add, {route, Route}) ->
```
michaeldjeffrey, line 241: Can we debounce this somehow?
We're going to slam the disk everytime we get a new connection or bring the box up.

src/grpc/iot_config/hpr_route_stream_worker.erl
```diff
@@ -253,39 +313,46 @@ maybe_init_from_file(Path) ->
...
-                Map
-            )
+                RoutesMap
+            ),
+            lists:foreach(fun hpr_route_ets:insert_eui_pair/1, EUIPairs),
+            lists:foreach(fun hpr_route_ets:insert_devaddr_range/1, DevAddrRanges)
     end.
 
--spec open_backup_file(Path :: string()) -> {ok, map()} | {error, any()}.
+-spec open_backup_file(Path :: string()) -> {ok, {map(), list(), list()}} | {error, any()}.
```
michaeldjeffrey, line 327: I'm wondering if there's still a benefit of having Routes in a map. 
We can `lists:usort` and keep everything as a list?

src/grpc/packet_router/hpr_packet_router_service.erl
```diff
@@ -38,6 +38,11 @@ route(EnvUp, StreamState) ->
                     ok = ?MODULE:register(PubKeyBin),
                     {ok, StreamState}
             end
+    catch
+        _E:_R ->
+            lager:warning("reason  ~p", [_R]),
+            lager:warning("bad envelope ~p", [EnvUp]),
+            {stop, StreamState}
```
michaeldjeffrey, line 45: Should we send back some sort of response?
This way we're not killing the connection and the gateway is endlessly reconnecting and sending bad packets.
Or does that matter?

michaeldjeffrey (dismissed):

src/grpc/packet_router/hpr_packet_router_service.erl
macpie: don't think it does

PR #175, opened 2023-04-12
Macpie/chatterbox

macpie:
This is only for testing DO NO MERGE


src/protocols/hpr_protocol_router.erl
```diff
@@ -192,6 +132,117 @@ terminate(_Reason, _State = #state{}) ->
...
+        [{_, #{channel := ChannelPid, stream_pid := StreamPid} = Stream}] ->
+            case erlang:is_process_alive(ChannelPid) andalso erlang:is_process_alive(StreamPid) of
+                true ->
+                    {ok, Stream};
+                false ->
+                    true = ets:delete(?STREAM_ETS, ?KEY(Gateway, LNS)),
+                    get_stream(Gateway, LNS, Server, BackoffFailed)
+            end;
+        [] ->
+            true = ets:insert(?STREAM_ETS, {?KEY(Gateway, LNS), ?CONNECTING}),
```
Vagabond, line 156: Use insert_new here and try again if it fails:

https://github.com/helium/blockchain-core/blob/master/src/ledger/v1/blockchain_ledger_v1.erl#L743

src/protocols/hpr_protocol_router.erl
```diff
@@ -192,6 +132,117 @@ terminate(_Reason, _State = #state{}) ->
...
+                                hpr_packet_router_downlink_handler,
+                                hpr_packet_router_downlink_handler:new_state(Gateway, LNS)
+                            }
+                        })
+                    of
+                        {error, _} = Error ->
+                            true = ets:delete(?STREAM_ETS, ?KEY(Gateway, LNS)),
+                            Error;
+                        {ok, Stream} ->
+                            true = ets:insert(?STREAM_ETS, {?KEY(Gateway, LNS), Stream}),
```
Vagabond, line 175: Use select_replace here:

https://github.com/helium/blockchain-core/blob/master/src/ledger/v1/blockchain_ledger_v1.erl#L808

src/protocols/hpr_protocol_router.erl
macpie: Thx, this should improve in case I get 2 at the exact same time, This is why I added some random jitter in test...

Vagabond, line 156: might be a good idea to use `{?CONNECTING, self()}` here

PR #186, opened 2023-05-05
set protocol version for http roaming to 1.0

michaeldjeffrey:
until we can become fully compliant with 1.1


macpie (approved):

macpie:
Not needed anymore

Issue #205, opened 2023-06-05
SKF usability


lytzV:
**Context**
LNS hosts may set up many routes all sharing the same DevAddr ranges. For instance, they could set up one route for each organization/tenant/user of their LNS, that way they can configure a max_copy setting on an organization/tenant/user level.

The current SKF works as following
• If a route has no SK on one of its DevAddr, then HPR will send all packets under that DevAddr to the route
• If a route has one or more SK on one of its DevAddr, then HPR will only send packets under that DevAddr if and only if HPR can validate the MIC using one of the SKs

Let's say an LNS host has 1000 organization/tenant/user, and 1000 DevAddr, neither of which are wildly unrealistic.
It is very much possible that each organization/tenant/user may not have an active device occupying each of the 1000 DevAddr, let's call those "inactive DevAddr." Note that one route's inactive DevAddr may be active in another route.

In order to avoid getting bombarded by traffic on inactive DevAddrs, LNS hosts may submit dummy keys on them, so HPR would fail at validating the MIC.

This means that the LNS host could be submitting 1000*1000 = 1mm SKs, despite the fact that many of them may simply be dummy keys.

This puts unnecessary burden on HPRs and config service.

**Proposal** 
Make a route-level flag "get all traffic for that route's DevAddr ranges"
• If the flag = true, then HPR would ignore all SKF on that route
• If the flag = false. 
    • When there is no SK on a DevAddr, instead of defaulting to "get all traffic for that DevAddr,"  default to "block all traffic for that DevAddr." That way, LNS hosts don't need to submit unnecessary dummy keys for inactive DevAddrs.
    • If a route has one or more SK on one of its DevAddr, then HPR will only send packets under that DevAddr if and only if HPR can validate the MIC using one of the SKs [unchanged from previous behavior]

lytzV:
Also need to think about how we facilitate the change to the new SKF scheme so we don't have a breaking change.

lytzV:
**Revised Proposal**
Make a route-level flag "ignore empty skf"

**If the flag = true**
• When there is no SK on a DevAddr, HPR blocks all traffic under that DevAddr to the route. That way, LNS hosts don't need to submit unnecessary dummy keys for inactive DevAddrs.

**If the flag = false**
• When there is no SK on a DevAddr, HPR sends all traffic under that DevAddr to the route.

[...]

Issue #201, opened 2023-05-30
Improve SKFs processing


macpie:
- [x] Use less memory
- [x] Improve processing time by having most used ones at the top
- [x] Avoid throwing data when not all SKFs are received yet
- [ ] Store on disk
- [ ] Optimize lookup further by some kind of ordering (look at fcnt)

PR #209, opened 2023-06-09
Macpie/skf optimize

macpie:
Fix for #205 #201


test/hpr_routing_SUITE.erl
```diff
@@ -1046,6 +1107,161 @@ maybe_report_packet_test(_Config) ->
...
+        max_copies => 1
+    }),
+    hpr_route_ets:insert_skf(SKF1),
+
+    {Time1, Result1} = timer:tc(hpr_routing, find_routes, [PacketType, PacketUp]),
+    ct:pal("[~p:~p:~p] MARKER ~p~n", [?MODULE, ?FUNCTION_NAME, ?LINE, {Time1, Result1}]),
+
+    [{_, SKFETS1}] = hpr_route_ets:lookup_route(Route1ID),
+
+    timer:sleep(2000),
```
michaeldjeffrey, line 1161: 2 second sleep?

test/hpr_routing_SUITE.erl
```diff
@@ -213,27 +197,99 @@ mic_check_test(_Config) ->
...
+        session_key => hpr_utils:bin_to_hex_string(NwkSessionKey),
+        max_copies => 3
+    }),
+    ?assertEqual(ok, hpr_route_ets:insert_skf(SKF)),
+
+    [{_, ETS}] = hpr_route_ets:lookup_route(RouteID),
+
+    %% Here we are making sure that the SKF got updated
+    [{_, BeforeUpdate, 3}] = hpr_route_ets:lookup_skf(ETS, DevAddr),
+    timer:sleep(2000),
```
michaeldjeffrey, line 284: 2 second sleep?

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -191,62 +217,105 @@ lookup_devaddr_range(DevAddr) ->
...
     ok.
 
--spec lookup_skf(DevAddr :: non_neg_integer()) -> [{binary(), string(), non_neg_integer()}].
-lookup_skf(DevAddr) ->
-    case ets:lookup(?ETS_SKFS, DevAddr) of
-        [] -> [];
-        SKFs -> [{Filter, RouteID, MaxCopies} || {_DevAddr, {Filter, RouteID, MaxCopies}} <- SKFs]
-    end.
+-spec lookup_skf(ETS :: ets:table(), DevAddr :: non_neg_integer()) ->
+    [{binary(), integer(), non_neg_integer()}].
```
michaeldjeffrey, line 293:
```suggestion
    [{SessionKey :: binary(), Timestamp :: integer(), MaxCopies :: non_neg_integer()}].
```
As a helper because I had a hard time reading the numbers swapped around.

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -74,18 +85,24 @@ delete_route(Route) ->
...
+            ets:delete(SKFETS);
+        _Other ->
+            lager:warning("failed to delete skf table ~p for ~s", [
+                _Other, RouteID
+            ])
+    end,
     true = ets:delete(?ETS_ROUTES, RouteID),
-    lager:info("deleted ~w DevAddr Entries, ~w EUIS Entries, ~w SKFs Entries for ~s", [
-        DevAddrEntries, EUISEntries, SKFsEntries, RouteID
+    lager:info("deleted ~w DevAddr Entries, ~w EUIS Entries for ~s", [
```
michaeldjeffrey, line 97: Do we not want to use the metadata anymore in logs? I liked that it made it easier to grep for messages.
```
lager:info(
  [{devaddr, DevAddrEntries}, {euis, EUIEntries}, {route_id, RouteID}],
  "route deleted"
),
```

src/hpr_routing.erl
```diff
@@ -7,9 +7,23 @@
     handle_packet/1
 ]).
 
+-ifdef(TEST).
+
+-export([
+    find_routes/2
```
michaeldjeffrey, line 13: Can we export this normally? 
I was having to reconstruct this function in the REPL the other day while I was investigating some routing.

src/hpr_routing.erl
```diff
@@ -77,87 +91,127 @@ find_routes({join_req, {AppEUI, DevEUI}}, _Packet) ->
...
+            case SelectedRoute of
+                undefined ->
+                    find_routes_for_uplink(
+                        Packet,
+                        DevAddr,
+                        Routes,
+                        EmptyRoutes,
+                        {Route, SessionKey, LastUsed, SKFMaxCopies}
+                    );
+                {_R, _S, L, _M} when LastUsed > L ->
```
michaeldjeffrey, line 163: This is...
We had a `SelectedRoute`, and found another applicable route. 
`SelectedRoute` get's swapped out if the new route has been used more recently.

We're not returning early when we find a first match, or delivering to multiple routes if for some reason they put the same (devaddr, session_key) in more than one place.

If the same (devaddr, session_key) is in more than one route, we could end up swapping where we send the packet every hour, which would be very confusing on many fronts.

test/hpr_routing_SUITE.erl
macpie: This is due to
 ```
-ifdef(TEST).
-define(SKF_UPDATE, timer:seconds(2)).
-else.
-define(SKF_UPDATE, timer:minutes(60)).
-endif.
```
 in `hpr_routing`

test/hpr_routing_SUITE.erl
macpie: same as above

src/grpc/iot_config/hpr_route_ets.erl
macpie: no you are right, fixing

src/hpr_routing.erl
macpie: We can

src/hpr_routing.erl
macpie: > We're not returning early when we find a first match, or delivering to multiple routes if for some reason they put the same (devaddr, session_key) in more than one place.

No because config service still allows same session keys in different routes so until we can address that we will have to check every route, I could pass the timestamp found to minimize select for other routes.

> If a (devaddr, session_key) pair was added to a different route, it would hijack the existing routing. And after a restart, routing would be dependent on db ordering from the config service.

Yes true, single session key will fix that.

michaeldjeffrey (approved):

PR #216, opened 2023-06-26
Improve multi buy to use ETS first to avoid too many calls to MBS

macpie:


src/hpr_multi_buy.erl
```diff
@@ -32,38 +32,30 @@ init() ->
...
+        LocalCounter ->
             case request(Key) of
-                {ok, C} when C > Max ->
+                {ok, ServiceCounter} when ServiceCounter > Max ->
+                    ets:update_counter(
+                        ?ETS,
+                        Key,
+                        {2, Max - LocalCounter + 1},
+                        {default, 0, erlang:system_time(millisecond)}
+                    ),
```
michaeldjeffrey, line 50:
```suggestion
                    ets:update_element(
                        ?ETS,
                        Key,
                        {2, ServiceCounter}
                    ),
```
`update_element/3` helps make it clearer that the next time around this packet will contain a value known to be higher than the max. And we don't have to do math 😄

macpie: `update_element/3`  is not `atomic and isolated`

michaeldjeffrey (approved):

PR #221, opened 2023-06-30
process leak

michaeldjeffrey:
- ~~link to packet delivery spawns.~~
- ~~update proto.~~
- move `grpcbox` to an `included_application`.
  - This keeps grpcbox from starting and accepting traffic until `hpr` is fully booted and decides to start `grpcbox` in `hpr_app.erl`.
 - Rework how router protocol monitors streams.
   - Monitor at the incoming stream level, when that goes down start a timer to remove all outgoing streams.
   - A gateway re-registering can cancel the timer and it will re-use the existing streams.
   - The server side of an outgoing stream can still close and remove a stream.

### DIFF
--- a/src/grpc/packet_router/hpr_packet_router_service.erl
+++ b/src/grpc/packet_router/hpr_packet_router_service.erl
@@ -33,12 +33,7 @@ route(eos, StreamState) ->
 route(EnvUp, StreamState) ->
     try hpr_envelope_up:data(EnvUp) of
         {packet, PacketUp} ->
-            _ = erlang:spawn_opt(
-                hpr_routing,
-                handle_packet,
-                [PacketUp],
-                [link, {fullsweep_after, 0}]
-            ),
+            _ = erlang:spawn_opt(hpr_routing, handle_packet, [PacketUp], [{fullsweep_after, 0}]),
             {ok, StreamState};
         {register, Reg} ->
             PubKeyBin = hpr_register:gateway(Reg),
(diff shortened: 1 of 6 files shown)


rebar.lock
```diff
@@ -59,7 +59,7 @@
  {<<"hackney">>,{pkg,<<"hackney">>,<<"1.17.0">>},0},
  {<<"helium_proto">>,
   {git,"https://github.com/helium/proto.git",
-       {ref,"c10a73fc89d895f439bc82efbbc8cd11ed880546"}},
+       {ref,"f1208e25b8966e0b480023439a53f5e159f9c0d0"}},
```
macpie, line 62: Why proto update?

src/grpc/packet_router/hpr_packet_router_service.erl
```diff
@@ -33,7 +33,12 @@ route(eos, StreamState) ->
 route(EnvUp, StreamState) ->
     try hpr_envelope_up:data(EnvUp) of
         {packet, PacketUp} ->
-            _ = erlang:spawn_opt(hpr_routing, handle_packet, [PacketUp], [{fullsweep_after, 0}]),
+            _ = erlang:spawn_opt(
+                hpr_routing,
+                handle_packet,
+                [PacketUp],
+                [link, {fullsweep_after, 0}]
```
macpie, line 40: Won't this crash the service if that packet crashes?

macpie:
Can you take a quick look at that failing `eunit` it seems to happen more often.

https://github.com/helium/helium-packet-router/actions/runs/5426366379/jobs/9869108656?pr=221

rebar.lock
michaeldjeffrey: I rebased onto master from when I started the branch and needed to update proto, this was the version that came along, I'll revert.

PR #241, opened 2023-07-31
Start work on session keys for data transfer.

ke6jjj:


macpie:
I ll build on top off this

src/grpc/packet_router/hpr_packet_router_service.erl
```diff
@@ -69,6 +64,10 @@ handle_info({give_away, NewPid, PubKeyBin}, StreamState) ->
     lager:info("give_away registration to ~p", [NewPid]),
     gproc:give_away({n, l, ?REG_KEY(PubKeyBin)}, NewPid),
     StreamState;
+handle_info(?SESSION_RESET, StreamState0) ->
+    {EnvDown, StreamState1} = create_session_offer(StreamState0),
+    %% TODO: Do we reset the session key here?
```
michaeldjeffrey, line 69: I'm not sure the order guarantees of bidirectional streams. 
But what I think this is asking is, when we send a new session offer, do we ignore any packets uplinked until the gw sends a new session key?
Probably a good gateway question, as they might change the order they need to process things in.

src/grpc/packet_router/hpr_packet_router_service.erl
```diff
@@ -114,6 +113,93 @@ register(PubKeyBin) ->
...
+                true ->
+                    lager:debug("session init ~s", [
+                        hpr_utils:bin_to_hex_string(hpr_session_init:nonce(SessionInit))
+                    ]),
+                    StreamState1 = grpcbox_stream:stream_handler_state(
+                        StreamState0, HandlerState0#handler_state{
+                            session_key = hpr_session_init:session_key(SessionInit)
+                        }
+                    ),
+                    ok = schedule_session_reset(),
```
michaeldjeffrey, line 193: If a gw sends a bunch of session_inits, it can start indefinitely many timers.

test/hpr_packet_router_service_SUITE.erl
```diff
@@ -0,0 +1,121 @@
...
+            route_id => RouteID, start_addr => 16#00000000, end_addr => 16#00000010
+        })
+    ],
+    {ok, GatewayPid} = hpr_test_gateway:start(#{
+        forward => self(), route => Route, eui_pairs => EUIPairs, devaddr_ranges => DevAddrRanges
+    }),
+
+    ok = hpr_test_gateway:send_packet(GatewayPid, #{}),
+
+    timer:sleep(1000),
```
michaeldjeffrey, line 85:
```suggestion
    %% Waiting session negotiation to complete
    timer:sleep(1000),
```

src/grpc/packet_router/hpr_packet_router_service.erl
madninja: > But what I think this is asking is, when we send a new session offer, do we ignore any packets uplinked until the gw sends a new session key?

You keep using the old/current session key (aka verifier key) until you get a new session init from the gateway

src/grpc/packet_router/hpr_packet_router_service.erl
macpie: will fix

michaeldjeffrey (dismissed):

PR #256, opened 2023-09-02
change how we store SKF for performance gains

michaeldjeffrey:
SKF are stored with the timestamp of the last time they were used to make it more likely to hit a session key for a packet early. An ets ordered_set table was used to achieve this, but it follows total sorting, which means the timestamp needs to be the first element in the table if we want that to take precedence.

Part of removing, or updating, a SKF involved `ets:select_delete/2` which requires a full table scan because we only had the SessionKey at the time of deletion. When maybe removes or updates come in from the config service, all of those updates would be spawned into their own process and contend for the SKF table at the same time. For sufficiently large ets tables (+10k) and sufficiently large batches of updates (1k), this would cause all schedulers to have trouble doing anything while tables scans were happening. This pausing would mean the grpc connection could not talk back to the config-service and negotiate a slower rate of updates, causing them to get out of order, and grpcbox would kill the connection. And it would all start again.

The structure of the table has been flattened, so we can tell ets the key position is the SessionKey in the second slot. It will still respect total ordering with the timestamp in the first position, but we can now do deletes in constant time with only the SessionKey. :dancer:

### DIFF
--- a/src/grpc/iot_config/hpr_route_ets.erl
+++ b/src/grpc/iot_config/hpr_route_ets.erl
@@ -141,13 +141,7 @@ insert_route(Route) ->
             [#hpr_route_ets{skf_ets = ETS}] ->
                 ETS;
             _Other ->
-                ets:new(?ETS_SKFS, [
-                    public,
-                    ordered_set,
-                    {read_concurrency, true},
-                    {write_concurrency, true},
-                    {heir, erlang:whereis(?SKF_HEIR), RouteID}
-                ])
+                mk_new_skf_ets(RouteID)
         end,
     ?MODULE:insert_route(Route, SKFETS).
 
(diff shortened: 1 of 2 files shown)


michaeldjeffrey:
The test data I was using mislead me. It also sorts on keypos. :sigh:

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -309,12 +303,9 @@ insert_skf(SKF) ->
     MD = skf_md(RouteID, SKF),
     case ?MODULE:lookup_route(RouteID) of
         [#hpr_route_ets{skf_ets = SKFETS}] ->
-            Deleted = do_remove_skf(SKFETS, SKF),
-            do_insert_skf(SKFETS, SKF),
-            case Deleted of
-                0 -> lager:debug(MD, "inserted SKF");
-                _ -> lager:debug([{deleted, Deleted} | MD], "updated SKF")
-            end;
+            true = do_remove_skf(SKFETS, SKF),
```
macpie, line 306: Can we make them return `ok` for both?

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -556,13 +542,25 @@ skfs_count_for_route(RouteID) ->
...
+%% on the second key.
+%% Structure: {Timestamp, SessionKey, DevAddr, MaxCopies}.
+-spec mk_new_skf_ets(hpr_route:id()) -> ets:table().
+mk_new_skf_ets(RouteID) ->
+    ets:new(?ETS_SKFS, [
+        public,
+        ordered_set,
+        {read_concurrency, true},
+        {write_concurrency, true},
+        {keypos, 2},
```
macpie, line 556: idk if this will work as the `ordered_set` will be on the key. Can you 2x check?

macpie: > one`` key, one object, ordered in Erlang term order

Maybe it's fine

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -572,7 +570,10 @@ do_insert_skf(SKFETS, SKF) ->
     %% This is negative to make newest time on top
     Now = erlang:system_time(millisecond) * -1,
     true = ets:insert(SKFETS, {
```
macpie, line 572: Maybe create a record for this?

PR #257, opened 2023-09-14
routing by packet lookup

michaeldjeffrey:
This PR changes the way packets get their routing determined, for the purpose of having more freedom in how we store SessionKey Filters.

### The when of packets
The first time a packet is see, we take a lock on the phash and start finding if it's routable.
Any packets that come in during that time will queue themselves.
Once the routing is figured out, the initial packet, and all the queued packets are routed, and the routing information is put into ETS.
Any packet coming along after routing has been determined will immediately retrieve it's routes.

Multi-buy is always checked.

### But there's a lot of changes to SKF storage...
The last_used timestamp for an SKF is removed. Worst case scenario is the applicable SKF is at the end of an ETS table, but that lookup is only done one time per hash.

This means we can keep SKF unique by session_key, and delete in constant time.

### Other things of note
- We no longer assume that if 2 routes have a valid SKF, the latest one wins. Now, both routes will be routed to.
- There's a cleanup job for phash routing. I based it loosely off of packet-purchaser unique packet hash counter. 
  - It crawls the table more often than the removal window to keep the table from getting bigger than can be traversed in a single window.
  - packet-purchaser was doing this while being a default router.
- There as a multi-buy test that was relying on a race condition. 
  - There was a max_copies of 3, and we were simulating 2 servers for every request. The test asserted that 3 requests were sent, but that could only happen if request 2 and 3 were sent at the same time. 
  - It's been updated to wait until multi-buy has received more than max_copies worth of function calls, and makes sure it's sent less than the expected requests.

### DIFF
--- a/src/grpc/iot_config/hpr_route_ets.erl
+++ b/src/grpc/iot_config/hpr_route_ets.erl
@@ -361,7 +361,7 @@ delete_skf(SKF) ->
             DevAddr = hpr_skf:devaddr(SKF),
             SessionKey = hpr_skf:session_key(SKF),
             MaxCopies = hpr_skf:max_copies(SKF),
-            %% Here we ignore max_copies
+
             true = ets:delete(SKFETS, hpr_utils:hex_to_bin(SessionKey)),
             lager:debug(
                 [
--- a/src/grpc/iot_config/hpr_route_ets.erl
+++ b/src/grpc/iot_config/hpr_route_ets.erl
@@ -23,7 +23,6 @@
     lookup_devaddr_range/1,
 
     insert_skf/1,
-    insert_new_skf/1,
     update_skf/4,
     delete_skf/1,
     lookup_skf/2,
@@ -317,21 +316,6 @@ insert_skf(SKF) ->
             lager:debug(MD, "updated SKF");
         _Other ->
             lager:error(MD, "failed to insert skf table not found ~p", [
-                _Other
-            ])
-    end,
-    ok.
-
--spec insert_new_skf(SKF :: hpr_skf:skf()) -> ok.
-insert_new_skf(SKF) ->
-    RouteID = hpr_skf:route_id(SKF),
-    MD = skf_md(RouteID, SKF),
-    case ?MODULE:lookup_route(RouteID) of
-        [#hpr_route_ets{skf_ets = SKFETS}] ->
-            do_insert_skf(SKFETS, SKF),
-            lager:debug(MD, "inserted SKF");
-        _Other ->
-            lager:error(MD, "failed to insert new skf, tabl not found ~p", [
                 _Other
             ])
     end,
--- a/src/grpc/iot_config/hpr_route_ets.erl
+++ b/src/grpc/iot_config/hpr_route_ets.erl
@@ -391,7 +391,8 @@ select_skf(Continuation) ->
     ets:select(Continuation).
 
 -spec select_skf(ETS :: ets:table(), DevAddr :: non_neg_integer() | ets:continuation()) ->
-    {[{SessionKey :: binary(), MaxCopies :: non_neg_integer()}], ets:continuation()} | '$end_of_table'.
+    {[{SessionKey :: binary(), MaxCopies :: non_neg_integer()}], ets:continuation()}
+    | '$end_of_table'.
 select_skf(ETS, DevAddr) ->
     MS = [{{'$2', {DevAddr, '$3'}}, [], [{{'$2', '$3'}}]}],
     ets:select(ETS, MS, 100).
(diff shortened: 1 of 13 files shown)


src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -382,9 +380,9 @@ delete_skf(SKF) ->
     ok.
 
 -spec lookup_skf(ETS :: ets:table(), DevAddr :: non_neg_integer()) ->
-    [{SessionKey :: binary(), Timestamp :: integer(), MaxCopies :: non_neg_integer()}].
+    [{SessionKey :: binary(), MaxCopies :: non_neg_integer()}].
 lookup_skf(ETS, DevAddr) ->
-    MS = [{{{'$1', '$2'}, {DevAddr, '$3'}}, [], [{{'$2', '$1', '$3'}}]}],
+    MS = [{{'$2', {DevAddr, '$3'}}, [], [{{'$2', '$3'}}]}],
```
jeffgrunewald, line 385: do these matchspec placeholders need to reset now that you're not returning three values?

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -393,9 +391,10 @@ select_skf(Continuation) ->
     ets:select(Continuation).
 
 -spec select_skf(ETS :: ets:table(), DevAddr :: non_neg_integer() | ets:continuation()) ->
-    {[{binary(), integer(), non_neg_integer()}], ets:continuation()} | '$end_of_table'.
+    {[{SessionKey :: binary(), MaxCopies :: non_neg_integer()}], ets:continuation()}
+    | '$end_of_table'.
 select_skf(ETS, DevAddr) ->
-    MS = [{{{'$1', '$2'}, {DevAddr, '$3'}}, [], [{{'$2', '$1', '$3'}}]}],
+    MS = [{{'$2', {DevAddr, '$3'}}, [], [{{'$2', '$3'}}]}],
```
jeffgrunewald, line 397: same question as above

src/hpr_routing.erl
```diff
@@ -97,95 +103,159 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
     ok = hpr_metrics:observe_find_routes(Time),
     Results.
 
+-spec start_crawl_routing() -> pid().
+start_crawl_routing() ->
```
jeffgrunewald, line 107: if this process dies is there anything that will re-start the recursion cycle or is it a problem if that _doesn't_ happen?

src/hpr_multi_buy.erl
```diff
@@ -70,6 +71,13 @@ cleanup(Duration) ->
     end),
     ok.
 
+-spec make_key(hpr_packet_up:packet(), hpr_route:route()) -> binary().
+make_key(PacketUp, Route) ->
+    crypto:hash(sha256, <<
+        (hpr_packet_up:phash(PacketUp))/binary,
```
jadeallenx, line 77: For the record, if this uses erlang:phash2 under the hood, you should know there is 1 in 27,000 chance of collision returning the same value. Not sure if you care about that - if this is a long lived look up value, I would find another way to get a more durable hash.

src/hpr_routing.erl
jadeallenx, line 107: Echo Jeff's comment - typically you would use a spawn_monitor here and if the pid dies, you'd get a message to this gen server to do something about it. If it's ok this dies and gets restarted randomly, maybe it doesn't matter.

src/hpr_routing.erl
```diff
@@ -97,95 +103,159 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
...
+    erlang:spawn(fun() ->
+        Timer = timer:seconds(hpr_utils:get_env_int(routing_cleanup_timer_secs, 10)),
+        Window = timer:seconds(hpr_utils:get_env_int(routing_cleanup_window_secs, 60)),
+        ok = timer:sleep(Timer),
+        Removed = do_crawl_routing(Window),
+        lager:debug(
+            [{timer, Timer}, {window, Window}, {removed, Removed}],
+            "cleaning routing table"
+        ),
+        true = erlang:unregister(?ROUTING_CLEANUP),
```
jadeallenx, line 117: Why are you registering a name here? You can carry the process pid around in your gen server state and fire off a cast or info message to send messages to it.

src/hpr_routing.erl
```diff
@@ -97,95 +103,159 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
...
+        true = erlang:register(?ROUTING_CLEANUP, start_crawl_routing())
+    end).
+
+-spec cancel_crawl_routing() -> ok.
+cancel_crawl_routing() ->
+    true = exit(get_crawl_routing_pid(), kill),
```
jadeallenx, line 123: Again, typically this would be a cast into gen_server state rather than an explicit message passed to the process

src/hpr_routing.erl
```diff
@@ -97,95 +103,159 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
...
+    ok.
+
+-spec get_crawl_routing_pid() -> undefined | pid().
+get_crawl_routing_pid() ->
```
jadeallenx, line 127: Same comment

src/hpr_routing.erl
michaeldjeffrey: This module is not a gen_server.

src/grpc/iot_config/hpr_route_ets.erl
michaeldjeffrey: They only need to match. But it's better to start them from 1.

src/hpr_routing.erl
michaeldjeffrey: The table belongs to the supervisor, if the table is not present this process will be reset.
But that leaves a chance that it comes back through the sleep, and now there's more crawls happening.
Which is not the worst since they're all doing the same sweep. But too many could bog down the table. 

hmmm, could probably spawn_link?
I'd have to rethink what it means to "cancel" the crawling so it doesn't kill the supervisor.
But it would keep the the supervisor restarting a bunch of times from making a fleet of these procs.

src/grpc/iot_config/hpr_route_ets.erl
```diff
@@ -362,17 +362,15 @@ delete_skf(SKF) ->
             SessionKey = hpr_skf:session_key(SKF),
             MaxCopies = hpr_skf:max_copies(SKF),
             %% Here we ignore max_copies
```
macpie, line 364:
```suggestion
```

src/hpr_routing.erl
```diff
@@ -50,45 +67,34 @@ handle_packet(PacketUp, Opts) ->
         {fun packet_throttle_check/1, [], packet_limit_exceeded}
     ],
     PacketUpType = hpr_packet_up:type(PacketUp),
-    case execute_checks(PacketUp, Checks) of
+    Check = execute_checks(PacketUp, Checks),
+    case Check of
```
macpie, line 71:
```suggestion
    case execute_checks(PacketUp, Checks) of
```

src/hpr_routing.erl
macpie: Yes, to echo Jeff & Jade's comments, I think  it might be easier to create a quick `gen_server` (`hpr_routing_cache` ?) that can handle that ETS table as well as the cleanup and all

macpie: I would also make it a record instead of a 3 tuple in the ETS, easier to read IMO.

src/hpr_routing.erl
```diff
@@ -97,95 +103,159 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
...
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+%% NOTE: all entries in `?PACKET_ETS' are 3-tuple {Hash, Time, Result} so it can
+%% be crawled with the same matchspec.
+-spec establish_routing(hpr_packet_up:packet(), Start :: non_neg_integer()) -> ok | {error, any()}.
+establish_routing(PacketUp, Start) ->
+    Hash = hpr_packet_up:phash(PacketUp),
+    true = ets:insert(?PACKET_ETS, {Hash, Start, {locked, [{PacketUp, Start}]}}),
```
macpie, line 146: If you really want to only do this once, I would do an insert new here, look at `hpr_protocol_router:292`

src/hpr_routing.erl
michaeldjeffrey: It's been moved to a gen_server `hpr_routing_cache`.

src/hpr_routing.erl
michaeldjeffrey: moved to `hpr_routing_cache` gen_server. and a record.

michaeldjeffrey:
The current largest DevAddr used has ~13.5k Session Key Filters.

Providing a packet that will never match any of the SKFs. (running with a single core locally).

| Number of SKfs | Worst case time |
|--------------|-------------------|
| 100 | 4ms |
| 1,000 | 27ms |
| 10,000 | 217ms |
| 20,000 | 431ms |
| 50,000 | 1027 ms |
| 100,000 | 2048ms |

src/hpr_routing.erl
```diff
@@ -101,91 +70,143 @@ find_routes({uplink, {_Type, DevAddr}}, PacketUp) ->
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+-spec do_handle_packet(hpr_packet_up:packet(), Start :: non_neg_integer()) -> ok | {error, any()}.
+do_handle_packet(PacketUp, Start) ->
+    Hash = hpr_packet_up:phash(PacketUp),
+    PacketUpType = hpr_packet_up:type(PacketUp),
+    case hpr_routing_cache:lookup(Hash) of
```
macpie, line 77:
```suggestion
    case ?MODULE:lookup(Hash) of
```

macpie (approved):

PR #259, opened 2023-09-25
Check that session init was done 5min after offer was sent

macpie:


test/hpr_packet_router_service_SUITE.erl
```diff
@@ -24,7 +25,8 @@
 %%--------------------------------------------------------------------
 all() ->
     [
-        session_test
+        session_test,
+        timeout_test
```
michaeldjeffrey, line 29: There's one more test we need.
resetting the session.

after the regular `register -> session_offer -> session_init -> ok`
we have the `?SESSION_TIMER` of 30 minutes. 
that restarts `session_offer -> session_init -> ok`. 

That is when we're concerned about the response timeout.
We send the 2nd session_offer, and if they don't respond to it, nothing happens.

The way I understand things in this PR, after 30 mins we send the session_offer, and then check 5 minutes later if the session_key exists. But we don't check if it was the same session_key that already existed.

src/grpc/packet_router/hpr_packet_router_service.erl
```diff
@@ -48,7 +48,7 @@ route(eos, StreamState) ->
     lager:debug("received eos for stream"),
     {stop, StreamState};
 route(EnvUp, StreamState) ->
-    lager:debug("got env up ~p", [EnvUp]),
+    ct:pal("got env up ~p", [EnvUp]),
```
michaeldjeffrey: lager

michaeldjeffrey (approved):

macpie:
Fixed in https://github.com/helium/helium-packet-router/pull/270

PR #265, opened 2023-10-24
Do not apply multi buy to join req anymore as they are free.

macpie:


src/hpr_routing.erl
```diff
@@ -8,6 +8,8 @@
     find_routes/2
 ]).
 
+-define(MAX_JOIN_REQ, 999).
```
macpie, line 11: @jthiller  @michaeldjeffrey This probably need to be discussed.

michaeldjeffrey (approved):

PR #296, opened 2024-02-26
Implement HIP 102

macpie:


config/ct.config
```diff
@@ -1,6 +1,8 @@
 [
     {hpr, [
         {data_dir, "data"},
+        %% See https://github.com/helium/HIP/blob/main/0102-helium-educational-lns.md
+        {hpr_free_net_ids, ["C00053"]},
```
michaeldjeffrey, line 5: If we put `16#C00053` in the settings, we can compare without needing to stringify.

michaeldjeffrey (approved):

config/ct.config
macpie: ok

michaeldjeffrey (approved):

PR #315, opened 2025-05-20
Add net ids stats

macpie:


src/cli/hpr_cli_info.erl
```diff
@@ -77,6 +79,28 @@ info_ips(["info", "ips"], [], []) ->
 info_ips(_, _, _) ->
     usage.
 
+info_netids(["info", "netids"], [], []) ->
+    List = lists:map(
+        fun({NetID, Count}) ->
+            #{
+                net_id => NetID,
```
michaeldjeffrey, line 86:
```suggestion
                net_id => hpr_utils:net_id_display(NetID),
```
So we don't have to convert between integer and hex after the fact?

michaeldjeffrey (approved):

PR #318, opened 2025-06-10
Bring back `hpr_timestamp_ms` as a binary

macpie:


macpie:
We realized that https://github.com/helium/helium-packet-router/blob/main/src/protocols/gwmp/hpr_gwmp_worker.erl#L277 already contains timestamp from HPR in `ms`

